PR #1021, opened 2015-10-31
Bug reporting steps

RobertoNovelo:
Since many issues are missing a common format, it is harder to review users reports. This changes shall make this process easier


FichteFoll:
I can assure you this won't help a bit. There is already a [`contributing.md`](https://github.com/wbond/package_control/blob/master/contributing.md), which is prominently referenced when you create an issue or even pull request. If people aren't respecting this, they won't be respecting this section in the readme below two licence boilerplate text sections.

Placing it at the top of the readme *may* help, but it may as well not.

Issue #935, opened 2015-04-21
Non-default channel does not work


solicomo:
I create a [private channel](https://raw.githubusercontent.com/solicomo/sublime-package-control-channel/master/channel.json) according to the [doc](https://packagecontrol.io/docs/channels_and_repositories), and imitating the [default channel](https://github.com/wbond/package_control_channel).

Then I added it to Package Control:

```json
{
	"channels":
	[
		"https://packagecontrol.io/channel_v3.json",
		"https://raw.githubusercontent.com/solicomo/sublime-package-control-channel/master/channel.json"
	],

	//...
}
```

But when I try to install package, the packages in my private channel do not show up.

wbond:
Please submit a debug log.

solicomo:
Here is the debug log:

```
Package Control: Fetching list of available packages
  Platform: windows-x64
  Sublime Text Version: 3083
  Package Control Version: 3.0.1
Package Control: Download Debug
  URL: https://packagecontrol.io/channel_v3.json
  Resolved IPv6: 2600:3c02::f03c:91ff:fedf:6883
  Resolved IP: 50.116.33.29
  Timeout: 30
Package Control: WinINet Debug Proxy
  proxy: 
  proxy bypass: 
  proxy username: 
  proxy password: 
Package Control: WinINet HTTPS Debug General
  Server SSL Certificate:
    subject: US, Massachusetts, Private Organization, 471714639, US, Massachusetts, Newbury, Codex Non Sufficit LC, codexns.io
    issuer: US, GeoTrust Inc., GeoTrust EV SSL CA - G4
    common name: codexns.io
    issue date: Wed, 17 Dec 2014 08:00:00 GMT
    expire date: Fri, 30 Oct 2015 07:59:59 GMT
Package Control: WinINet HTTPS Debug Write
  GET /channel_v3.json HTTP/1.1
  Accept-Encoding: bzip2,gzip,deflate
  If-Modified-Since: Wed, 22 Apr 2015 03:00:05 GMT
  If-None-Match: "55370eb5-32e6e"
  User-Agent: Package Control v3.0.1
  Host: packagecontrol.io
  Connection: Keep-Alive
  Cache-Control: no-cache
Package Control: WinINet HTTPS Debug Read
  HTTP/1.1 200 OK
  Server: nginx
  Date: Wed, 22 Apr 2015 03:23:34 GMT
  Content-Type: application/json
  Content-Length: 206082
[...]

PR #1030, opened 2015-11-20
Mark the packages_cache as required.

dwarburt:
Currently, without the packages_cache in the channel json, no packages can be discovered.  This has something to do with the way the PackageManager builds its list of repositories.  In the [list_repositories method](https://github.com/wbond/package_control/blob/818b598a0e0259b791cbee638de2575ef2653195/package_control/package_manager.py#L383) it checks in the package_cache for info about the "releases" for a repository and if they are not found then the package goes into the unavailable packages list and doesn't become available to install. 

I verified this claim by forking channel_v3.json in gist and removing the package_cache and replacing my main feed with that one and it failed to discover any repositories to install.  [forked channel here](https://gist.githubusercontent.com/anonymous/2a60a2c19742d2157cfb/raw/979cd64b3e47a4ba78f5cebafbb96b9cfc8d5416/channel_v3.json).

I think either this change should be made to the channel documentation or a fix to the PackageManager, but this change is it only one I have time for now so proposing this.  If you think the PackageManager should be fixed instead (or if I misunderstood something) then please reject this request.

### DIFF
--- a/example-channel.json
+++ b/example-channel.json
@@ -12,7 +12,7 @@
 		"https://github.com/SublimeText"
 	],
 
-	// The "packages_cache" is completely optional, but allows the
+	// The "packages_cache" is required and allows the
 	// channel to cache and deliver package data from multiple
 	// repositories in a single HTTP request, allowing for significantly
 	// improved performance.


FichteFoll:
Was this not fixed by 2066136adfd6d7d00a958cb58b7a13aaa1d15a72 (#935)? Either way, since this is a bug it should be fixed in the code and not documented to have bugged behavior, imo.

wbond:
I tested the forked channel and it functioned properly with the latest version of Package Control in the master branch.

I do appreciate your time in trying to help out. Sorry it took so long to get to reviewing this PR.

PR #1081, opened 2016-02-13
Workaround to prevent issue 1002

jrnewell:
Workaround to prevent issue 1002 while waiting for @wbond to push new trust_list code.

Ignore 'The specified item could not be found in the keychain' error reported by OS X command '/usr/bin/security dump-trust-settings -d'

PR #1121, opened 2016-06-26
Use oscrypto to generate the CA bundles

ThinkChaos:
I only tested this on OS X, but it should work on Windows too thanks to your work in oscrypto.

The one thing I'm not sure about and that should be done before merging is how to properly add oscrypto as a dependency.
Do I just add it in dependencies.json next to bz2 ([here](https://github.com/wbond/package_control/blob/5643f4fa9a335a4591482a8ecdf6e0d7413fbaa1/dependencies.json#L17))?

PS: If you merge this, you should probably close #1081.

### DIFF
--- a/package_control/ca_certs.py
+++ b/package_control/ca_certs.py
@@ -1,10 +1,10 @@
 import os
-import oscrypto
 import time
 import ssl
 import sys
 
 from .console_write import console_write
+from .deps import oscrypto
 from .open_compat import open_compat, read_compat
 
 try:
--- a/package_control/ca_certs.py
+++ b/package_control/ca_certs.py
@@ -4,7 +4,7 @@ import ssl
 import sys
 
 from .console_write import console_write
-from .deps import oscrypto
+from .deps.oscrypto import trust_list
 from .open_compat import open_compat, read_compat
 
 try:
@@ -217,5 +217,5 @@ def create_ca_bundle(settings, destination):
     """
 
     with open_compat(destination, 'w') as f:
-        for der_bytes, _, _ in oscrypto.trust_list.extract_from_system():
+        for der_bytes, _, _ in trust_list.extract_from_system():
             f.write(ssl.DER_cert_to_PEM_cert(der_bytes))
--- a/package_control/ca_certs.py
+++ b/package_control/ca_certs.py
@@ -1,22 +1,11 @@
+import base64
 import os
-import re
+import oscrypto
 import time
 import sys
-import struct
-import locale  # To prevent import errors in thread with datetime
-import datetime
-import base64
 
-if os.name == 'nt':
-    import ctypes
-    from ctypes import windll, wintypes, POINTER, Structure, GetLastError, FormatError
-    crypt32 = windll.crypt32
-
-from .cmd import Cli
 from .console_write import console_write
 from .open_compat import open_compat, read_compat
-from .unicode import unicode_from_os
-from .http.x509 import parse_subject, parse
 
 try:
     str_cls = unicode
@@ -141,22 +130,21 @@ def get_system_ca_bundle_path(settings):
             is_old = stats.st_mtime < time.time() - 604800
 
         if not exists or is_empty or is_old:
-            if platform == 'darwin':
-                if debug:
+            if debug:
+                if platform == 'darwin':
                     console_write(
                         u'''
                         Generating new CA bundle from system keychain
                         '''
                     )
-                _osx_create_ca_bundle(settings, ca_path)
-            elif platform == 'win32':
-                if debug:
+                elif platform == 'win32':
                     console_write(
                         u'''
                         Generating new CA bundle from system certificate store
                         '''
                     )
-                _win_create_ca_bundle(settings, ca_path)
+
+            create_ca_bundle(settings, ca_path)
 
             if debug:
                 console_write(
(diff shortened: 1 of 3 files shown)


package_control/ca_certs.py
```diff
@@ -141,22 +130,21 @@ def get_system_ca_bundle_path(settings):
             is_old = stats.st_mtime < time.time() - 604800
 
         if not exists or is_empty or is_old:
-            if platform == 'darwin':
-                if debug:
+            if debug:
+                if platform == 'darwin':
```
FichteFoll: I suppose these platform-specific debug messages are redundant now.

ThinkChaos: Yeah that's another thing I wasn't sure if I should just merge, but decided not to as they still hold, even though there's an extra layer of abstraction.
@wbond What do you think?

wbond:
I don't want oscrypto as a real dependency for two reasons:

 1. It is just Python, not shared libraries 
 2. Maintaining hashes in the PC code base to verify the authenticity of the dependency would be a hack.

There is another part of Package Control that will benefit from asn1crypto, and that is the UrllibDownloader. Currently it has a hack of an ASN.1 library that needs to be replaced.

The most likely solution will be a dev script that extracts the necessary parts of oscrypto and asn1crypto and convert the imports to be relative.

I'm not sure if I want to keep the current one week cache of the trust roots. From testing on my 2013 rMBP, exporting all of the trust roots and encoding them to a file takes around 200ms. That shouldn't be too bad if it is part of the background downloader thread and we do it every 24 hours.

In general Windows has far fewer trust roots than OS X, and Linux already has them in the correct format.

ThinkChaos:
How about using git submodules, and either relative imports, or `sys.path` hacks?

> From testing on my 2013 rMBP, exporting all of the trust roots and encoding them to a file takes around 200ms.

On my mid 2012 MBP it takes more than that:

    $ python -m timeit -s 'from oscrypto.trust_list import extract_from_system' 'extract_from_system()'
    10 loops, best of 3: 3.13 sec per loop

I do have a lot of custom policies in my keychain though, which might cause a slowdown.

wbond:
> How about using git submodules, and either relative imports, or sys.path hacks?

I don't need any of the symmetric encryption aspects of oscrypto, but even withstanding that, I would want to do them as relative imports. Package Control should be 100% self-contained. Recent changes to ST3 including bundling `bz2` and `ssl` on Linux, so those are no longer required to be bootstrapped. I would consider it a failure to move backwards and add a new dependency on something that can be easily embedded. `sys.path` hacks are hacks, and lead to more hacks to deal with module unloading, reloading and other stuff like that. In an single-namespace world like Python, the most sane way to deal with dependencies is to merge them into the package in question. Then you don't have to deal with the backflips of version compatibility, synchronized upgrades, etc. It is Python's version of static linking. :smile:

I really have no problem with copying code. That is how [packagecontrol.io pulls in parts of Package Control](https://github.com/wbond/packagecontrol.io/blob/master/setup/scripts/extract_package_control.py). submodules just complicate usage of the repository.

I don't see this as a problem that needs a different solution, just something that needs to be done.

> On my mid 2012 MBP it takes more than that:

The export requires parsing every certificate in your trust root. On my machine that is 193 certificates. If you have significantly more, I could see that being a linear slowdown. You can also see a 5-15% difference in execution time based on what version of Python the code is being run on. Python 2 is slower than Python 3.

Either way, that should happen in a background thread, so if for some users it takes 20 seconds, their updates will just take a little longer to install.

ThinkChaos:
How about writing a Python script that uses pip to install oscrypto locally?
It could also remove all files we don't care about after pip is done. But I don't think this is the best idea as you'd have to track internal imports in oscrypto and its dependencies. Which doesn't take too long, but is a bit annoying and error prone. Do you think the space savings would be worth it?

Also, to make sure, this script would be run in the dev repo, and never on the end user side. Right?


I ran the timeit benchmark using Python 3.5.1, and have 178 root certificates, every single one of which has custom trust settings.
Anyways I agree this could easily be done in a background thread. So not a big issue.

Maybe we could only remove the caching if the parsing takes more than x seconds?
This would be a different pull request though, as this one should probably be hurried a bit since it fixes a user facing error.

Sorry for taking so long to respond, I wasn't getting my email notifications!

wbond:
Since I'm the author of oscrypto and asn1crypto, I'm happy to deal with tracking what is needed by Package Control. Those packages were written to deal with Package Control and some other projects I have built. I'm not 100% sure I'll trim them down, but at the very least I want the import to be relative.

ThinkChaos:
The script ([get_deps](https://github.com/ThinkChaos/package_control/blob/master/get_deps)) is almost there.
I have a couple things I'd like your opinion on:
- How do I prevent Sublime from thinking the script is a plugin?  
  As a workaround I removed the .py extension.
- How do you think I should patch the imports in the dependencies?
  If this were a shell script I would've made a diff, and used `patch`, which isn't a nice solution in Python.
  Here are the imports that require changes: [oscrypto/trust_list.py](https://github.com/wbond/oscrypto/blob/41cb25705aabe33003c89dcc307c81cb23ee650b/oscrypto/_linux_bsd/trust_list.py#L6-L7).

ThinkChaos:
bump

wbond:
@ThinkChaos Thank you for your work on this.

I ended up making some changes to asn1crypto and oscrypto to make the integration easier. I also didn't want to rely on pip being installed to copy the dependencies over, and I needed to rewrite the content of the oscrypto files to use asn1crypto via relative imports.

Because of this and a few other issues, I ended up writing be747b94c778cc8d9815746aa3bec1ad3eea247f from scratch. I certainly don't mean to discount your work by not using your commits directly.

To answer your question about loading a `.py` file as a plugin – that only happens if the `.py` file is in the root of the package. I worked around this by placing it in a folder named `dev/`.

ThinkChaos:
No worries, I'm just glad this is fixed.
I'll just be waiting on the release now!

> I worked around this by placing it in a folder named `dev/`.

Of course!

PR #1158, opened 2016-09-21
Added the setting detection for config.sublime-package-override

evandrocoan:
When the exists the file `config.sublime-package-override`, the
package control attempts to install the package to `Installed
Packages`, instead of mark the package as installed.

See the issue: #1155

### DIFF
--- a/package_control/package_cleanup.py
+++ b/package_control/package_cleanup.py
@@ -153,7 +153,8 @@ class PackageCleanup(threading.Thread):
             clean_old_files(package_dir)
 
             if int(sublime.version()) > 3000 and os.path.exists(package_dir + '/.sublime-package-override'):
-                found = False
+                if not os.path.exists(package_dir + '/.no-sublime-package'):
+                    found = False
 
             # Cleanup packages/dependencies that could not be removed due to in-use files
             cleanup_file = os.path.join(package_dir, 'package-control.cleanup')
--- a/package_control/package_cleanup.py
+++ b/package_control/package_cleanup.py
@@ -152,7 +152,7 @@ class PackageCleanup(threading.Thread):
 
             clean_old_files(package_dir)
 
-            if os.path.exists(package_dir + '/config.sublime-package-override' ):
+            if os.path.exists(package_dir + '/.sublime-package-override' ):
                 found = False
 
             # Cleanup packages/dependencies that could not be removed due to in-use files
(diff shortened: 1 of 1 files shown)


wbond:
This only handles automatic installation when packages are present in `installed_packages` and not as a `.sublime-package`.

This would also need to be handled in the `package_control/package_manager.py`. It should also be behind version guards (overrides only exist for ST3). This way a user can have overrides in place, and then use the *Install Package* command as expected.

Finally, I think the filename should be `.sublime-package-override` for consistency with `.no-sublime-package` and `.sublime-dependency`.

FichteFoll:
An override package is also mutually exclusive to a package specifying `.no-sublime-package`, which doesn't appear to be checked here either.

evandrocoan:
> This only handles automatic installation when packages are present in installed_packages and not as a .sublime-package.

I could not understand what you mean here. The only thing which need to be done is a simple feature. When the user create a unpacked package at his folder `SublimeText\Data\Packages` (for portable ST), and there is the file `.sublime-package-override` on the package folder... The Package Control must ensures the package is installed at the package control packages files at `SublimeText\Data\Installed Packages` (for portable ST). So, because of this, I could not understand what do you mean by.

May be I got a hint, are you saying that on the `SublimeText\Data\Packages` (for portable ST), could be sublime-text packages files as `Google Spell Check.sublime-package`?

On this case, I am not catching them, but it is not the reason for this feature. If the user has the file as `Google Spell Check.sublime-package`, just unzip it and put it a his folder `SublimeText\Data\Packages` (for portable ST), or just replaces or add if not yet, the package to ``SublimeText\Data\Installed Packages` (for portable ST)`.

But if you think the zipped package case should be handled, I may implement it. But I just tested putting a package as `Google Spell Check.sublime-package` at the `SublimeText\Data\Packages` (for portable ST), and at the ST launch, the Package Control started installing it as it was not installed. So, the package manager was already installing the packages on the user folder `SublimeText\Data\Packages` (for portable ST), to the `SublimeText\Data\Installed Packages` (for portable ST) folder, despite there is a `.sublime-package-override` or not on the zipped package file.

>An override package is also mutually exclusive to a package specifying .no-sublime-package, which doesn't appear to be checked here either.

Thanks. When a package specifies `.no-sublime-package` it is not a package,
and there is not sense to apply `.sublime-package-override`, so
just in case this happens, it will be ignored. Was this you meant?

evandrocoan:
I am closing it because I use my branch named `master` to do this path, but now i want to do other things with my branch named `master`. So, I am opening it on an branch with a proper name for this feature.

PR #1236, opened 2017-06-15
Sublime package override feature

evandrocoan:
This continues:

1. https://github.com/wbond/package_control/pull/1158 Added the setting detection for config.sublime-package-override

Because I had the misfortune to use the `master` branch to do the pull request. Now this pull request is on a proper branch.


deathaxe:
Package Control 4.0 will install all missing packages, even if overrides already exist.

PR #1240, opened 2017-06-16
Error message flood fix

evandrocoan:
When there is a batch performance for some chain of execution, and there are error on them,
we cannot show a error message dialog for each one of them immediately, otherwise we would
flood the user with messages. 

See:

1. https://github.com/SublimeTextIssues/Core/issues/1510 Sublime Text Files syntax error flooding

### DIFF
--- a/package_control/show_error.py
+++ b/package_control/show_error.py
@@ -1,4 +1,10 @@
-import sublime
+
+try:
+    # Allow using this file on the website where the sublime
+    # module is unavailable
+    import sublime
+except (ImportError):
+    sublime = None
 
 from . import text
 from .console_write import console_write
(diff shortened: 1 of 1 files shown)


package_control/show_error.py
```diff
@@ -21,5 +34,31 @@ def show_error(string, params=None, strip=True, indent=None):
...
+                    If there will be new error messages on the next seconds,
+                    they will be show on the Sublime Text console
+                '''
+                ) )
+
+            is_error_recentely_displayed = True
+
+            # Enable the message dialog after x.x seconds
+            thread = Timer(60.0, _restart_error_messages)
+            thread.start()
```
deathaxe: I am wondering if this can be handled using `sublime.set_timeout()` here?

evandrocoan: Sure, but I do not what is the difference. Usually I prefer to avoid sublime API, because with my experience it seems more bugged than the Python.

deathaxe:
PC 4.0 will take another route to solve this issue.

Issue #1238, opened 2017-06-16
AttributeError: 'bool' object has no attribute 'replace'


evandrocoan:
I am using the dev version on this commit:

1. https://github.com/wbond/package_control/commit/b4cb03d2457bcac29ba3cfc13689098aee103971

I just did a full clean install of everything, and on the end of it all, it keeps doing this error:
```
Exception in thread Thread-26:
Traceback (most recent call last):
  File "./python3.3/threading.py", line 901, in _bootstrap_inner
  File "D:\SublimeText\Data\Packages\Package Control\package_control\automatic_upgrader.py", line 117, in run
    self.upgrade_packages()
  File "D:\SublimeText\Data\Packages\Package Control\package_control\automatic_upgrader.py", line 257, in upgrade_packages
    ignore_packages=self.auto_upgrade_ignore
  File "D:\SublimeText\Data\Packages\Package Control\package_control\package_installer.py", line 99, in make_package_list
    incoming = upgrader.incoming()
  File "D:\SublimeText\Data\Packages\Package Control\package_control\upgraders\git_upgrader.py", line 107, in incoming
    info = self.get_working_copy_info()
  File "D:\SublimeText\Data\Packages\Package Control\package_control\upgraders\git_upgrader.py", line 59, in get_working_copy_info
    branch = res.replace('refs/heads/', '')
AttributeError: 'bool' object has no attribute 'replace'
```

Full debug log: 

1. [Package Control: Debug for AttributeError: 'bool' object has no attribute 'replace'](https://gist.github.com/evandrocoan/757da5a3c7083b26c6069dc3f85f5980)

This is my settings file:

1. https://github.com/evandrocoan/SublimeTextStudio/blob/4346adf573308be7eef05d7795372ed39567880e/User/Package%20Control.sublime-settings

wbond:
This should be resolved by 790bcd6

Issue #1233, opened 2017-05-24
Why ignore VCS-based packages accordingly to this message?


evandrocoan:
This error pops up some times I open Sublime Text. 

I just enabled debug mode, when find the log I post it here.

```
Package Control

Error executing: "C:\Program Files (x86)\Git\cmd\git.exe" symbolic-ref -q HEAD

Working directory:
D:\SublimeText\Data\Packages\PowerCursors

VCS-based packages can be ignored by changing the "ignore_vcs_packages" setting to true.
Sublime Text will need to be restarted once the setting is changed.
```

evandrocoan:
I recorded the log with debug on:

1. [[LOG] Why ignore VCS-based packages accordingly to this message?](https://gist.github.com/evandrocoan/8b24910c8a9a24066941219d5259d20a)

```
[...]

PR #1241, opened 2017-06-16
Symbolic ref head fix

evandrocoan:
Fix issues:

1. https://github.com/wbond/package_control/issues/1233 Why ignore VCS-based packages accordingly to this message?
1. https://github.com/wbond/package_control/issues/1238 AttributeError: 'bool' object has no attribute 'replace'


wbond:
I ended up fixing this in 790bcd6 – thanks for taking the time to report the source of the issue.

PR #1251, opened 2017-07-31
Fix attempted relative import in non-package when importing it

evandrocoan:
outside Sublime Text.

https://stackoverflow.com/questions/33837717/systemerror-parent-module-not-loaded-cannot-perform-relative-import

### DIFF
--- a/package_control/cmd.py
+++ b/package_control/cmd.py
@@ -3,10 +3,16 @@ import subprocess
 import re
 import sys
 
-from .console_write import console_write
-from .unicode import unicode_from_os
-from .show_error import show_error
-from . import text
+try:
+    from .console_write import console_write
+    from .unicode import unicode_from_os
+    from .show_error import show_error
+    from . import text
+except:
+    from console_write import console_write
+    from unicode import unicode_from_os
+    from show_error import show_error
+    import text
 
 if os.name == 'nt':
     from ctypes import windll, create_unicode_buffer
--- a/package_control/console_write.py
+++ b/package_control/console_write.py
@@ -7,8 +7,10 @@ except (NameError):
     # Python 3
     str_cls = str
 
-from . import text
-
+try:
+    from . import text
+except:
+    import text
 
 def console_write(string, params=None, strip=True, indent=None, prefix=True):
     """
--- a/package_control/show_error.py
+++ b/package_control/show_error.py
@@ -1,6 +1,9 @@
 import sublime
 
-from . import text
+try:
+    from . import text
+except:
+    import text
 
 
 def show_error(string, params=None, strip=True, indent=None):


wbond:
This doesn't seem to solve an issue with Package Control. What are you trying to accomplish?

evandrocoan:
I am building a script which can be run from a Sublime Text Package, but also from the command line as well. But when I tried to run it from the command line, it does not accepted the relative imports.  This is the script:

1. https://github.com/evandrocoan/SublimeTextStudioChannel/blob/0f5f0f2ba573a82eb4b1cd70c85dc672245a1b21/update.py#L68-L80

wbond:
If you import `package_control`, then all relative imports should work properly. This is how the loader exposes the `package_control` module within Sublime Text and how packagecontrol.io re-uses the downloader/provider code.

evandrocoan:
If I remove this patch, when I attempt to import it:
```python
assert_path( os.path.join( os.path.dirname( current_directory ), "Package Control" ) )

# https://stackoverflow.com/questions/9123517/how-do-you-import-a-file-in-python-with-spaces-in-the-name
package_control = importlib.import_module( "package_control" )
from package_control import cmd
```

I throws this error:
```
  File "update.py", line 81, in <module>
    from package_control import cmd
  File "/cygdrive/d/SublimeText/Data/Packages/Package Control/package_control/cmd.py", line 6, in <module>
    from console_write import console_write
ModuleNotFoundError: No module named 'console_write'
```

wbond:
Try:

```python
import sys
import os

sys.path.append(os.path.join(os.path.dirname(current_directory), "Package Control"))

from package_control import cmd
```

evandrocoan:
Still not working. I printed the path for you to see:
```python
def print_python_envinronment():
    index = 0;

    for path in sys.path:
        print(index, path);
        index += 1;

current_directory = os.path.dirname( os.path.realpath( __file__ ) )
assert_path( os.path.join( os.path.dirname( current_directory ), 'PythonDebugTools' ) )
sys.path.append( os.path.join( os.path.dirname( current_directory ), "Package Control" ) )

print_python_envinronment()
from package_control import cmd
```
-->
```
python3 update.py
0 /cygdrive/d/SublimeText/Data/Packages/PythonDebugTools
1 /cygdrive/d/SublimeText/Data/Packages/StudioChannel
2 /usr/lib/python36.zip
3 /usr/lib/python3.6
4 /usr/lib/python3.6/lib-dynload
5 /usr/lib/python3.6/site-packages
6 /cygdrive/d/Data/Packages/Package Control
  File "update.py", line 75, in <module>
    from package_control import cmd
  File "/cygdrive/d/User/Dropbox/Applications/SoftwareVersioning/SublimeText/Data/Packages/Package Control/package_control/cmd.py", line 6, in <module>
    from console_write import console_write
ModuleNotFoundError: No module named 'console_write'

```

wbond:
Line 6 should not be `from console_write import console_write`. I think you forgot to revert a change. It is `from .console_write import console_write`.

evandrocoan:
You are right, but I did not forget to revert, I actually reverted to the wrong code. It worked just by changing these lines:
```python
assert_path( os.path.join( os.path.dirname( current_directory ), "Package Control" ) )
cmd = importlib.import_module("cmd")
```
-->
```python
assert_path( os.path.join( os.path.dirname( current_directory ), "Package Control" ) )
from package_control import cmd
```

PR #1277, opened 2017-10-04
Update asn1crypto to 0.23.0

dsanders11:


dsanders11:
@wbond, I threw in changing the version to 3.3.1 as well to try to entice you to merge this ASAP and get it released. If there's anything else that speeds up that process, let me know.

wbond:
Been busy today cutting, priming and painting doors. 😀

A pre-release will always need to go out before a new version. Don't want to bork PC for millions of users, so we need to test it with the small subset who have opted in to prereleases.

dsanders11:
@wbond, well is a pre-release anything I can move towards on my end? Seems like something you'd have to do, so I guess we're back to you finding the time.

dsanders11:
@wbond, bump.

deathaxe:
See 3350b43d3cdf17e8344ad6dcaa81fea462c889b2 how to mark a version as beta.

dsanders11:
@deathaxe, I've changed it to be '3.3.1-beta1' instead but this is just shuffling papers unless someone has the ability to actually do a pre-release.

deathaxe:
The basic question is whether an ordinary PR should include a version bump. Maybe other things need/should be included first before doing so.

dsanders11:
@deathaxe, I only included it to try to speed up the process since it seems like @wbond has no spare time. It can easily be removed.

But I would like to see some movement on this, it's starting to take an annoyingly long amount of time considering the trivial nature of the fix and the fact that it affects anyone who has installed OpenVPN (not a small number of people).

dsanders11:
@wbond, I'm going to keep bugging you until this gets done or you allow someone else to put out a release. There are now multiple issues on the repo from people running into this problem, and surely hundreds more in the wild. It's frustrating and disheartening that the fix is available and nothing is being done.

wbond:
@dsanders11 You are kind of a jerk. At the end of the day my primary responsibility is to my family and not some self-entitled coder on the internet who has the tools to solve the issue himself (and already has).

PR #1280, opened 2017-10-05
Fixed trying to exhaustively install a missing dependency

evandrocoan:
When the dependency is locally cloned on the loose Packages folder as
a git repository and the setting "ignore_vcs_packages" is set to
true, and would trigger the error message:

Skipping git package PythonDebugTools since the setting
"ignore_vcs_packages" is set to true

This happens when you develop the dependency on you
computer and the dependency is installed by git.


deathaxe:
No longer relevant for Package Control 4

PR #1286, opened 2017-10-14
Fix skip cleanup user dir

ndunks:
My custom plugin or menu that written in Package/User directory always deleted when package_cleanup is running because is not listed as 'installed', I write code to fix by skipping 'User' package/dir.
I'm using windows 10 x64 and SublimeText 3

### DIFF
--- a/package_control/package_cleanup.py
+++ b/package_control/package_cleanup.py
@@ -1,6 +1,7 @@
 import threading
 import os
 import functools
+import sys
 
 import sublime
 
@@ -148,6 +149,11 @@ class PackageCleanup(threading.Thread):
 
             package_dir = os.path.join(sublime.packages_path(), package_name)
             if not os.path.isdir(package_dir):
+                continue
+
+            # Skipping user package
+            if package_name == "User":
+                console_write("Skip cleanup " + package_dir)
                 continue
 
             clean_old_files(package_dir)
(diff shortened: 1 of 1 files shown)


package_control/package_cleanup.py
```diff
@@ -150,6 +150,11 @@ def run(self):
             if not os.path.isdir(package_dir):
                 continue
 
+            # Skipping user package
+            if package_name == "User":
+                console_write("Skip cleanup " + package_dir)
```
deathaxe: Not sure whether it's worth a message in console as it is always the User package which is skipped. This is nothing to worry or pay attention to in some way.

FichteFoll:
You shouldn't have a `package-control.cleanup` file in your user package though, which appears to be the only situation where this would matter.

wbond:
It is probably reasonable to skip the `User` and `Default` packages. However, as @FichteFoll mentioned, there should never be any of the private Package Control marker files in those dirs anyway.

Maybe there *should* be a message in the log telling users they are doing something very wrong?

deathaxe:
I recently found Package Control to turn a manually created package into a PC controlled one by adding the meta-data when running the "Upgrade Overwrite all" command and a package of that name exists in the repo. `Default` and `User` should never be part of the public repository, but maybe there are some edge cases with private ones, which would end up in that issue? 

The one way or the other - both packages should be made sure to never be handled by PC in any way, I think.

deathaxe:
The guard has been added to e7d7eaa821f429660a9414772d6716976b89fb26 and will be part of PC4.0

PR #1289, opened 2017-10-22
Use sublime.cache_path() to store cache data

### LINKED ISSUE #1268: Use sublime.cache_path() to store cache data
We talked about this before, but having an issue about this should help not forgetting about it.

(Just stumbled upon it again while preparing backups.)

deathaxe:
closes #1268

This commit moves all 'state' files to Sublime Text's cache folder.

TODO:
1. How to efficiently migrate/cleanup the old folders and files?
2. Keep ST2 compatibility as it doesn't know about `cache_path()`

### DIFF
--- a/package_control/automatic_upgrader.py
+++ b/package_control/automatic_upgrader.py
@@ -64,7 +64,11 @@ class AutomaticUpgrader(threading.Thread):
 
         self.last_run = None
 
-        self.last_run_file = os.path.join(sublime.packages_path(), 'User', 'Package Control.last-run')
+        cache_path = os.path.join(sublime.cache_path(), 'Package Control')
+        if not os.path.isdir(cache_path):
+            os.mkdir(cache_path)
+
+        self.last_run_file = os.path.join(cache_path, 'last-run')
 
         if os.path.isfile(self.last_run_file):
             with open_compat(self.last_run_file) as fobj:
--- a/package_control/automatic_upgrader.py
+++ b/package_control/automatic_upgrader.py
@@ -64,11 +64,15 @@ class AutomaticUpgrader(threading.Thread):
 
         self.last_run = None
 
-        cache_path = os.path.join(sublime.cache_path(), 'Package Control')
+        if int(sublime.version()) < 3000:
+            cache_path = os.path.join(sublime.packages_path(), 'User')
+        else:
+            cache_path = os.path.join(sublime.cache_path(), 'Package Control')
+
         if not os.path.isdir(cache_path):
             os.mkdir(cache_path)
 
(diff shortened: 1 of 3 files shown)


FichteFoll:
With the release of 3.0, Will has decided to drop ST2 support.

package_control/automatic_upgrader.py
```diff
@@ -64,7 +64,15 @@ def load_last_run(self):
...
-        self.last_run_file = os.path.join(sublime.packages_path(), 'User', 'Package Control.last-run')
+        if int(sublime.version()) < 3000:
+            cache_path = os.path.join(sublime.packages_path(), 'User')
+        else:
+            cache_path = os.path.join(sublime.cache_path(), 'Package Control')
+
+        if not os.path.isdir(cache_path):
+            os.mkdir(cache_path)
+
+        self.last_run_file = os.path.join(cache_path, 'Package Control.last-run')
```
FichteFoll: Might as well remove the "Package Control" part from the file name itself.

package_control/ca_certs.py
```diff
@@ -213,19 +213,23 @@ def get_system_ca_bundle_path(settings):
...
 
     # If the sublime module is available, we bind this value at run time
-    # since the sublime.packages_path() is not available at import time
+    # since the sublime.cache_path() is not available at import time
     global ca_bundle_dir
 
     if not ca_bundle_dir:
-        ca_bundle_dir = os.path.join(sublime.packages_path(), 'User')
-    if not os.path.exists(ca_bundle_dir):
+        if int(sublime.version()) < 3000:
```
FichteFoll: I don't know about Will's policies here, but since this code is being reused here, it might make sense to outsource it as `get_cache_path` (or similar) into a shared module.

package_control/automatic_upgrader.py
deathaxe: I readded it because of `cache_path` pointing to User package on ST2. Would need to built the file name depending on ST version, too.

package_control/ca_certs.py
deathaxe: Possible, but didn't want to add further modules in the first step. Efficiently it is required this way in 2 places, only, so didn't find it worth creating a shared module for a single function. http_cache.py needs some little different processing to ensure nested folders are created well.

wbond:
I would like to fix a couple of small bugs (a minor release I am working on now) and then make the break with ST2

deathaxe:
Ok, so feel free to just pick what you need or find better solutions.

wbond:
Sorry if I was unclear, I am planning to include this, I am just going to wait for the break with ST2. :smile:

deathaxe:
"include" includes "pick what you need" 😆

PR #1290, opened 2017-10-22
Avoid duplicated file exists checks

deathaxe:
The `open_compat` function calls `os.path.exists()` if a file is opened for reading and raises `FileNotFoundError` in case the file does not exist.

Therefore external calls to `os.path.exists` cause a file to be checked twice if it exists.

This commit therefore encapsulates all reading operations into `try..except` statements and catches the `FileNotFoundError` to handle it if required.

Furthermore this commit fixes an issue with file pointer to `messages.json` not being closed if `ValueError` is raised while parsing.


wbond:
Thanks!

PR #1305, opened 2017-12-21
Fix: No error for invalid dependencies not displayed

deathaxe:
If invalid dependencies are found, they are not added to the list of `invalid_dependencies` and therefore no error message is created.


wbond:
Thanks!

PR #1306, opened 2017-12-22
Avoid duplicated directory exists checks.

deathaxe:
The `os.makedirs()` function raises an OSError by default if the path already exists.
By passing `exist_ok=True` the function fails silently in such cases and can therefore be directly used to optionally create directories.


wbond:
This looks to be 3.3+, so I'll have to wait until ST2 is dropped

PR #1307, opened 2017-12-22
Some improvements of the cache module

deathaxe:
- fix a typo
- clear the internal cache dictionary with its `clear()` method
- avoid creating empty `list` and `dict` objects as default values, which are not used


wbond:
Thanks!

PR #1308, opened 2017-12-22
Optimize file extension recognition in 'clear_directory' module

deathaxe:
This commit makes use of the builtin `str.endswith()` method to

- replace fnmatch() function
- replace string splicing


wbond:
Great improvements, thanks!

PR #1309, opened 2017-12-22
Fix: header parsing fails if wget fails to download the content

deathaxe:
If wget doesn't exit cleanly, the `header_lines` list generated in `parse_output()` can be empty, which causes `parse_headers()` to try to read from `self.tmp_file` again. As it was deleted by `parse_output()` already, an `FileNotFoundException` is raised but not handled anywhere, causing any following process to fail.

Reading from tmp_file is therefore removed from `parse_header()` method.


wbond:
LGTM

PR #1359, opened 2018-05-10
Output Package Control Messages without empty whitespace

dnicolson:
This pull request changes the way "Package Control Messages" are displayed, it avoids outputting whitespace when displaying blank newlines.

If you are using the [Trailing Spaces](https://github.com/SublimeText/TrailingSpaces) package, you will see this:

screen shot 2018-05-10 at 23 52 41

### DIFF
--- a/package_control/package_manager.py
+++ b/package_control/package_manager.py
@@ -1754,7 +1754,10 @@ class PackageManager():
 
         def read_message(message_path):
             with open_compat(message_path, 'r') as f:
-                return '\n  %s\n' % read_compat(f).rstrip().replace('\n', '\n  ')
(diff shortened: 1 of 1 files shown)


wbond:
I believe it is reading one *message file* at a time, not *one line* at a time.

dnicolson:
> This seems like a good idea, but it isn't clear to me how this code will fix the problem in the graphic.
> 
> Wouldn't it need to replace every instance of `\n \n` with `\n\n`?

It's been a while, you're absolutely right... 😆

dnicolson:
Revised in https://github.com/wbond/package_control/pull/1450.

PR #1361, opened 2018-05-21
Add export package list functionality

Grundligt:
Added a command that creates a new file with a comma-separated list of the installed packages.


FichteFoll:
Not a bad idea.

Issue #1391, opened 2018-11-21
Dependency dependencies


Thom1729:
Currently, packages may have dependencies, but dependencies may not themselves have dependencies. This means that dependencies must either vendor the code they depend on (partly defeating the purpose of having dependencies) or advise packages to declare extra dependencies. In the latter case, the dependency author must also declare the correct loading order.

Instead, dependencies could declare other dependencies in the same way as packages. Then, load order can be determined by a topological sort. (To ensure backward compatibility, PC should load dependencies with a declared loading order in that order, then load all other dependencies in topological order.)

The advantages of this approach are:

- Dependencies never have to vendor code that's already available as another dependency.
- A dependency adding code that relies on another dependency is no longer a breaking change.
- Dependency authors no longer have to manually determine load order. A change in load order is no longer breaking.
- Packages only have to declare dependencies that they use directly, not dependencies of dependencies.
- It should be safe to load dependencies as soon as they are installed (without restarting Sublime).

Thom1729:
I should mention that I'm willing to implement this myself; I'd just like to get some feedback before spending the time to do so.

FichteFoll:
I believe there is some more important stuff to do first, like moving dependencies to the Lib folder and extracting them properly or getting rid of st2 code paths, but this is generally much appreciated from my side.

Thom1729:
I think that these things are orthogonal. This would be a fairly small code change; much, much smaller than the ST2 deprecation (#1313) and likely smaller than moving dependencies (although I'm not clear on exactly how that would work). I don't think there's any need to do one before the others.

FichteFoll:
True enough.

PC does need some work, but implementing dependency dependencies is probably an easy addition on top of the current features and imo definitely worthwhile to implement *and* support.

PR #1392, opened 2018-11-27
Allow dependencies to specify their own dependencies.

Thom1729:
For #1391.

The core change here is so small and self-contained that I think it can/should be done before worrying about broader issues like automatic load order or moving the dependency directory. While this change doesn't fix everything, it will be useful right away.

### DIFF
--- a/package_control/package_manager.py
+++ b/package_control/package_manager.py
@@ -1047,12 +1047,11 @@ class PackageManager():
         release = packages[package_name]['releases'][0]
 
         have_installed_dependencies = False
-        if not is_dependency:
-            dependencies = release.get('dependencies', [])
-            if dependencies:
-                if not self.install_dependencies(dependencies):
-                    return False
-                have_installed_dependencies = True
+        dependencies = release.get('dependencies', [])
+        if dependencies:
+            if not self.install_dependencies(dependencies):
+                return False
+            have_installed_dependencies = True
 
         url = release['url']
         package_filename = package_name + '.sublime-package'


FichteFoll:
Looks simple enough. I also believe these are the only locations where dependencies matter, so we are *probably* good regarding installation. However, what about satisfying dependencies? Idk if dependencies are skipped there.

I wonder whether adding an example with channel-specified dependencies to `example-repository.json` would be worth it, however. I don't think anyone uses this feature currently and while it could be used to resolve dependencies or do batch installations, PC's installation code (as you can see) only looks at it when the package has been downloaded anyway.

Thom1729:
The Satisfy Dependencies command uses `find_required_dependencies()`, so this patch should work with that.

It looks like exactly two packages specify dependencies in the channel. I wonder if we should just deprecate the feature and fix those two packages. This PR should work either way, though.

```python
from package_control.package_manager import PackageManager
[
    (package_name, release["dependencies"])
    for package_name, package in PackageManager().list_available_packages().items()
    for release in package["releases"] if "dependencies" in release
]
```

Thom1729:
Incidentally, both of the two packages that use channel dependencies (FileBrowser and LaTeXing) also have a `dependencies.json` file (which is presumably being ignored).

FichteFoll:
That would be a topic for a separate issue, though. Since afaik the schema follows semver it'd require dumping it to 4.0.0, but that's the least of our problems.

FichteFoll:
Oh, that reminds me, did you test this?

Thom1729:
Briefly, but I'm sure I didn't hit all of the corner cases. For one thing, there aren't any dependencies out there with dependencies.

evandrocoan:
> Thom1729, For one thing, there aren't any dependencies out there with dependencies

There is this dependency with dependencies, the `portalockerfile`: (requires the dependency `python-pywin32`)
1. https://github.com/evandroforks/portalockerfile

Next, the dependency `concurrentloghandler`: (requires the dependency `portalockerfile`)
1. https://github.com/evandroforks/concurrentloghandler

Next, the dependency `debugtools`: (requires (optionally) the dependency `concurrentloghandler`)
1. https://github.com/evandrocoan/debugtools

i.e., when loading the `debugtools` dependency, it detects whether `portalockerfile` is available. If so, then, it uses it. Otherwise, it fall back to a built-in behavior.

Currently, It just asks to the users require the required dependencies chain at once, when including the `debugtools` dependency.
```javascript
{
    "windows": {
        ">3000": [
            "python-pywin32",
            "portalockerfile",
            "concurrentloghandler",
            "debugtools"
        ]
    },
    "*": {
        "*": [
            "portalockerfile",
            "concurrentloghandler",
            "debugtools"
        ]
    }
}
```

But with the feature of dependencies requiring dependencies, this should be reduced to:
```javascript
{
    "*": {
        "*": [
            "concurrentloghandler",
            "debugtools"
        ]
    }
}
```

Because the `concurrentloghandler` is an optional dependency.

evandrocoan:
@Thom1729, I think it is missing to patch the `PackageManager.get_dependencies(package_name)` method to recursively get the nested dependencies. I just saw it being used here:
1. https://github.com/SublimeText/UnitTesting/commit/38f90ddf5bb0545cdd7232f1731c435d51886ac0#diff-5198be055987db9f3c6d0b91637897ffR119

Thom1729:
I don't think we can do that here. Specifically, `PackageManager.get_dependencies(package_name)` can't get transitive dependencies because it's possible that the immediate dependencies may not be loaded, so we can't necessarily get their dependencies.

That package reloading code can compute the transitive closure, though, because it's safe to assume that all of the dependencies are actually loaded. So I think that the reloading code should do this itself. (I actually wrote that code a month or so ago, so I can fix it up.)

Thom1729:
I've submitted a PR to solve the recursive reload problem in [AutomaticPackageReloader](https://github.com/randy3k/AutomaticPackageReloader) (which is functionally upstream from UnitTesting).

In principle, `PackageManager` could provide a helper method to recursively find all dependents of a dependency. In practice, I doubt that very many packages would use such a thing, and I'm leery of adding anything to the undocumented, not-officially-public API.

evandrocoan:
> I'm leery of adding anything to the undocumented, not-officially-public API.

It is not undocumented. It has a documentation right in the top of it. It just do not get a fancy webpage as the ones generated by sphinx or doxygen:
```python
File: package_control/package_manager.py
191:     def get_dependencies(self, package):
192:         """
193:         Returns a list of dependencies for the specified package on the
194:         current machine
195: 
196:         :param package:
197:             The name of the package
198: 
199:         :return:
200:             A list of dependency names
201:         """
```
It says it returns the list of dependencies for a packages. As it does not specify which dependencies, it should be implicit to assume it returns all dependencies, instead random counts as 1, 2, half, first level, etc.

___
> I've submitted a PR to solve the recursive reload problem in AutomaticPackageReloader

If PackageManager `get_dependencies` already did that, it would not be required to change the implementation for `AutomaticPackageReloader`. 

Now, the next package which requires to know all dependencies of a package will be required to copy-paste your implementation from `AutomaticPackageReloader`. 

If the implementation was performed directly on PackageManager `get_dependencies`, a copy-and-paste would not be required.

___
> can't get transitive dependencies because it's possible that the immediate dependencies may not be loaded, so we can't necessarily get their dependencies.

If they are not loaded? I think you meant installed. Then, they should be:
1. Installed before continuing/returning from `get_dependencies`
1. Or raising an exception of missing dependencies.
1. Or halting the recursive computation and assuming an empty result for the nested dependencies.

I still did not understand why you say it will not break things. Not implementing to recursively getting dependencies will break the behavior for `get_dependencies`. 

Follow the example from before and after this pull request gets merged:
1. If a package A requires the dependency B, which requires the dependency C
   1. With the current implementation on `master`:
      1. The package A needs to directly require the dependencies B and C
      1. Then, a call to `get_dependencies` will return the dependencies B and C
   1. After this get merged into master:
      1. The package A now can only require the dependency B (as now, the dependency C will be required by the dependency B)
      1. Then, a call to `get_dependencies` will return only the dependency B (instead of B and C as it did before this pull request gets merged into master)

Thom1729:
Incidentally, the AutomaticPackageReloader PR is not getting dependencies -- it's doing the opposite, getting dependents. This could be implemented in terms of a recursive dependency getter method, but that would be very inefficient. (We'd have to map it over all dependencies and packages, and each iteration would call the expensive `list_packages` and `list_dependencies` methods.) Even if we modified `get_dependencies` to do a recursive search, we would need a new, separate method to avoid duplicating AutomaticPackageReloader's functionality.

A lot of AutomaticPackageReloader's code is duplicated by other packages that reload modules. Ideally, we should factor that code out into a dependency. I would like to do that soon-ish, but there are other improvements I'd like to make to that code before inviting other packages to depend on it.

---

I concede that there's an argument for providing this functionality in Package Control. I don't agree that it's appropriate to change `get_dependencies`.

In the first place, PackageManager uses `get_dependencies` internally. It relies on its present functionality. If we changed `get_dependencies`, we'd have to preserve the old functionality in a new method. Either way, we're talking about adding a method with the new functionality; the only question is whether the new or old method gets the name `get_dependencies`.

Currently, `get_dependencies` does a simple thing with a simple implementation that requires few assumptions. A recursive implementation does something more complicated and requires more assumptions. These additional assumptions are a burden upon the caller.

> If they are not loaded? I think you meant installed. Then, they should be:
> 
> 1. Installed before continuing/returning from get_dependencies

I think that this is a non-starter. We can't block the caller while waiting to install something from the internet.

> 2. Or raising an exception of missing dependencies.

This makes sense in principle, but other PC "internals" methods do not do this. We would be adding a new failure mode that did not exist elsewhere. This would also be a breaking change.

> 3. Or halting the recursive computation and assuming an empty result for the nested dependencies.

This is the best option, but the caller would need to be aware that the information returned may not be correct. At that point, there's not much gained by keeping the same name.

---

Hooking into Package Control internals is inherently hazardous. It's not something that most packages would ever need or want to do, and a package that does is usually doing something fairly technical. AutomaticPackageReloader is a good example of this. Package Control provides a low-level interface that allows other packages to navigate those hazards in a way that works for their specific use cases. If we change `get_dependencies` to do more work, then we might be saving someone a few lines of code, but we might also be laying a trap by hiding complicated behavior and unobvious assumptions beneath a simple-looking API.

For these reasons, I don't think that we should change the behavior of the existing `get_dependencies` method. It may be useful to provide additional utility methods for recursively computing dependencies or dependents, but I don't think that those features are essential for this PR.

evandrocoan:
> > If they are not loaded? I think you meant installed. Then, they should be:
> > 1. Installed before continuing/returning from get_dependencies

> I think that this is a non-starter. We can't block the caller while waiting to install something from the internet.

We can. I did it a few times, check an example:
```python
# some other code before...
can_continue  = [False]
active_window = sublime.active_window()

# This run on Sublime Text thread
def on_done(item_index):
   # do some logic
   can_continue[0] = True

def show_quick_panel(selected_index=0):
   active_window.show_quick_panel( informations,
           on_done, sublime.KEEP_OPEN_ON_FOCUS_LOST, selected_index )

# show_quick_panel is a non-blocking function, but I need it to be a blocking function
show_quick_panel()

# this is my main thread, it waits until the show_quick_panel callback running on 
# another thread signalizes it has finished
while not can_continue[0]:
   time.sleep( 1 )
```
https://github.com/evandrocoan/channelmanager/blob/434c28f6bfc221357ebd4526c6d2737a3ddbaa0b/all/channel_manager/channel_installer.py#L1766


> If we change get_dependencies to do more work, then we might be saving someone a few lines of code

Not changing the current behavior of `get_dependencies`, will cause `get_dependencies` to do less work than it currently does, because it will start reporting a package has less dependencies than it actually has. As I already exemplified on the last post. A package nowadays calls `get_dependencies` and they will correctly get all dependencies for that package, because Package Control nowadays does not support dependencies with dependencies, then, all dependencies which have dependencies nowadays, have to be explicitly included in the package requiring the main/first dependency (which causes `get_dependencies` to correctly report 100% all dependencies, including nested dependencies). In the future, after this gets merged, when a package calls `get_dependencies`, it will only get the first level dependencies, because now, the dependencies' dependencies are required per dependency and `get_dependencies` is not recursively calculating the package dependencies, i.e., the dependencies' dependencies will not be anymore all required at once by the package requiring the main/first dependency, because this is done per dependency as Package Control supports it. `(this paragraph is hard to understand, you need to pay attention to every word and commas)`

> I concede that there's an argument for providing this functionality in Package Control. I don't agree that it's appropriate to change get_dependencies.

The last paragraph before this quote should explain why the behavior of `get_dependencies` needs changing after merging this pull request. If this still not clear, please, ask for clarifications in what troubles you. 


> > 2. Or raising an exception of missing dependencies.

> This makes sense in principle, but other PC "internals" methods do not do this. We would be adding a new failure mode that did not exist elsewhere. This would also be a breaking change.

Indeed, you are right. My idea was just a dumb ideia.

> > 3. Or halting the recursive computation and assuming an empty result for the nested dependencies.

> This is the best option, but the caller would need to be aware that the information returned may not be correct. At that point, there's not much gained by keeping the same name.

This is the second best option, as it minimizes the change this pull request will create. If this option is not implemented, `get_dependencies` will fail 100% of the cases to get all dependencies for a package. But, if this option is implemented, most times `get_dependencies` can correctly report all dependencies required by a package. Just remembering, nowadays, without this pull request being merged, calling `get_dependencies` will report 100% of the time, correctly 100% of all dependencies for a package, because packages will are required to require not only the direct dependencies, but also all dependencies's dependencies.

The first and best option to keep the current behavior of `get_dependencies`, after this pull request is merged, is the first option I mentioned, and it can be implemented with the trick I mentioned on the beginning of this post.

> Incidentally, the AutomaticPackageReloader PR is not getting dependencies -- it's doing the opposite, getting dependents. 

I do not understood what this means exactly. I would have to look in to AutomaticPackageReloader code to understand better.


> This could be implemented in terms of a recursive dependency getter method, but that would be very inefficient. (We'd have to map it over all dependencies and packages, and each iteration would call the expensive `list_packages` and `list_dependencies` methods.) Even if we modified `get_dependencies` to do a recursive search, we would need a new, separate method to avoid duplicating AutomaticPackageReloader's functionality.

This I think I understood. That is probably the implementation I am thinking of and AutomaticPackageReloader, as you explained, seems to be doing better in performance efficiency.

> A lot of AutomaticPackageReloader's code is duplicated by other packages that reload modules.

How they dare to do such thing?

> I concede that there's an argument for providing this functionality in Package Control

Package Control already get this functionality due how things are implemented, i.e., not supporting dependencies requiring dependencies. At least, the behavior of `get_dependencies` is consistent.

> In the first place, PackageManager uses `get_dependencies` internally. It relies on its present functionality. If we changed `get_dependencies`, we'd have to preserve the old functionality in a new method. Either way, we're talking about adding a method with the new functionality; the only question is whether the new or old method gets the name `get_dependencies`.

If you understood what I had explained, you will see a new method is not required.

> Currently, `get_dependencies` does a simple thing with a simple implementation that requires few assumptions. A recursive implementation does something more complicated and requires more assumptions. These additional assumptions are a burden upon the caller.

Indeed, the current implementation of `get_dependencies` is very simple, and the new implementation I am ranting all about will be extremely large, a monster implementation, because it will require to handle several things, i.e., edge cases.

> For these reasons, I don't think that we should change the behavior of the existing get_dependencies method. It may be useful to provide additional utility methods for recursively computing dependencies or dependents, but I don't think that those features are essential for this PR.

You are working on your free time and good will, I would not expect anything more from you, other and thank you for your contributions for the Sublime Text Packages community. Whether the monster I am suggesting to `get_dependencies` became, is implemented by you on this pull request or by someone else in another pull request, is irrelevant. For me, I think you could merge this pull request and the very aggressive behavior changes I am suggesting for `get_dependencies` cold be implemented either by me, or anyone else which gets troubled by the changes this will cause to `get_dependencies`. All I am stating, will all these long post lines, is that someday, someone could get in trouble by the changes to `get_dependencies` this pull requesting is causing. These posts are only for the sake explaining that things are not going to be 100% backward compatible, and someday, someone may get in trouble.

> Hooking into Package Control internals is inherently hazardous.

Package Control just downloads some stuff from internet, unzip them, and let Sublime Text load them. It is not some hacky package which does very nasty stuff to be considered hazardous. Hazardous, is a very strong word. I would call hazardous a package which mock with Windows API using ctypes, as touching anything relating to M\*i--c\*r-o%s¨o\_f=t is very dangerous.

> we might be saving someone a few lines of code, but we might also be laying a trap by hiding complicated behavior and unobvious assumptions beneath a simple-looking API.

Indeed, `get_dependencies` will be more than a simple method, and a big note should be put in its documentation explaining everything it might be doing when someone is calling it.

Thom1729:
> We can. I did it a few times, check an example:

I mean “can't” in the sense of “absolutely must not”. Users could be calling `get_dependencies()` from the main thread. Calling `sleep` on the main thread freezes the whole program. This would be an egregious breaking change.

This leaves the following possible behaviors for `get_dependencies('MyPackage')`:

1. Return exactly the immediate dependencies of MyPackage.
2. Return an unknown subset of the recursive dependencies of MyPackage depending on the transient state of the Package Control internals.

> I do not understood what this means exactly.

In AutomaticPackageReloader, when we reload MyPackage, we don't care what MyPackage depends on, but we do care what depends on MyPackage. It's the inverse relationship.

> How they dare to do such thing?

Um, because more than one package needed the same logic. We've discussed consolidating the logic into a dependency. I imagine that we probably will do this.

The point is that there's no point in worrying about this one piece of logic being duplicated when that piece is part of a large chunk that is duplicated, and we're not going to implement the whole chunk in Package Control anyway. The right thing to do is to factor out that whole chunk into a dependency — at which point the recursive dependent logic is no longer duplicated, resolving the original objection.

> Package Control just downloads some stuff from internet, unzip them, and let Sublime Text load them.

This is a gross oversimplication. In particular, “let Sublime Text load them” encompasses a great deal of complex logic. Python was never designed for the kind of dynamic module loading that Sublime's package system requires. I do suggest taking a look at AutomaticPackageReloader.

But more germanely, “downloads some stuff from internet” has a great number of possible failure modes, and Package Control has to account for and gracefully recover from each one. This is not the kind of logic that we can reasonably insert into an existing method that currently just reads a list from a file and returns the result.

evandrocoan:
> I mean “can't” in the sense of “absolutely must not”. Users could be calling get_dependencies() from the main thread. Calling sleep on the main thread freezes the whole program. This would be an egregious breaking change.

I am absolutely not suggesting anyone to call `sleep()` from the main thread. And if you pay attention to the simplified example (or the full linked code), you will see there are two threads mentioned. The Sublime Text main thread, which signalizes some boolean value, and my main program custom thread, which is waiting (calling `sleep()`) for the Sublime Text main thread give the white card.

> Return an unknown subset of the recursive dependencies of MyPackage depending on the transient state of the Package Control internals.

If I am not mistaken, this should be the option I first presented, and you pointed as the best solution, and the option I latter pointed as the second best solution. But indeed, after think about this better, I also would call this to be the best solution because:
1. The special case when the user call `get_dependencies` and not all dependencies are available, should be very rare.
1. If the user is worried get `get_dependencies` does not return fully all dependencies (because not all of them are installed yet), he should look for some way to ensure everything is installed. ``(just remember, the dependency recursive calculation/checking is a file system search, reading `dependencies.json` files, it does not require dependencies to be loaded on `sys.path` or anything else by python)``
   ```python
   def get_dependencies(self, package):
       """
       Returns a list of dependencies for the specified package on the
   
       current machine
       :param package:
           The name of the package
   
       :return:
           A list of dependency names
       """
   
       if package_file_exists(package, 'dependencies.json'):
           dep_info_json = read_package_file(package, 'dependencies.json')
           if dep_info_json:
               try:
                   return self.select_dependencies(json.loads(dep_info_json))
               except (ValueError):
                   console_write(
                       u'''
                       An error occurred while trying to parse the
                       dependencies.json for %s.
                       ''',
                       package
                   )
   
       metadata = self.get_metadata(package)
       if metadata:
           return metadata.get('dependencies', [])
   
       return []
   ```
Then, a simple explanation note on the `get_dependencies` function documentation, should be more than enough for this much simpler implementation.

> In particular, “let Sublime Text load them” encompasses a great deal of complex logic. Python was never designed for the kind of dynamic module loading that Sublime's package system requires. I do suggest taking a look at AutomaticPackageReloader.

Indeed, Python was not designed for such hacks implemented by Sublime Text to create a dynamic plugin environment. But this complex logic dealing with the inner workings of Python is not implemented by Package Control, but Sublime Text and AutomaticPackageReloader because they directly work with these details.

I just quickly checked the code looking for the `package_control` usage. And the only thing I found was the usage for listing available packages and dependencies and checking whether something is a dependency or not.

Package Control has, or at least try to have some logic protecting the Sublime Text user when the does some stuff as uninstalling his current color scheme. Package Control ensures the `color_scheme` setting is properly set, so the User does not get a error from Sublime Text, complaining about missing `color_scheme`.

> But more germanely, “downloads some stuff from internet” has a great number of possible failure modes, and Package Control has to account for and gracefully recover from each one. This is not the kind of logic that we can reasonably insert into an existing method that currently just reads a list from a file and returns the result.

Indeed, Package Control is a very robust/well implemented software. And when it downloads some stuff from the internet, there countless problems/errors scenarios already though beforehand to ensure Package Control works most of the time, if not always and by any means, as even working with proxy user connections.

> > How they dare to do such thing?

> Um, because more than one package needed the same logic...

I was just joking when I said "dare". Of course, there are countless reasons for duplicating some code or logic.

Thom1729:
> I am absolutely not suggesting anyone to call sleep() from the main thread. And if you pay attention to the simplified example (or the full linked code), you will see there are two threads mentioned. The Sublime Text main thread, which signalizes some boolean value, and my main program custom thread, which is waiting (calling sleep()) for the Sublime Text main thread give the white card.

Users may already be calling `get_dependencies()` on the main thread.

---

I think at this point we've established our opinions.

evandrocoan:
> Users may already be calling get_dependencies() on the main thread.

On this case then, it should not be a problem because the more reliable solution concluded was to just stop when it cannot find the dependencies installed.

> > 3. Or halting the recursive computation and assuming an empty result for the nested dependencies.

PR #1393, opened 2018-12-09
Allowed the install_package_command.py to select/install several packages at once

evandrocoan:
fsg

Related:

1. https://github.com/skuroda/PackageResourceViewer/pull/21 Augmented extract packages

### DIFF
--- a/package_control/commands/install_package_command.py
+++ b/package_control/commands/install_package_command.py
@@ -3,6 +3,8 @@ import threading
 import sublime
 import sublime_plugin
 
+from .advanced_install_package_command import AdvancedInstallPackageThread
+
 from .. import text
 from ..show_quick_panel import show_quick_panel
 from ..package_installer import PackageInstaller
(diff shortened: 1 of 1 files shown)


FichteFoll:
Interesting idea. Haven't looked at the code yet, however.

Can you show how this looks when filtering and then selecting? Do you select the first item to submit your selection?

I suggest listing the selected packages in the first item. I would also remove the `(excluded)` tags and only mention when a package is selected.

evandrocoan:
I implemented this initially 364 days ago for my dependency:
1. https://github.com/evandrocoan/channelmanager/commit/e75b6ee54bc1fe0a09153981c9da669024da07ca Allowed the command Select Package to Update Git Tag to accept

After this first implementation, took me a a lot of time testing and refining it. Then, after a year using it, I though it would be nice if I could also use this feature while installing packages, and everywhere else I could select one thing or more.

> Can you show how this looks when filtering and then selecting? Do you select the first item to submit your selection?

It works like this:
1. First you select a item
1. Then, this item is placed as the top item with the ` (selected)` tag.
1. Then, 
   1. You can hit the start item to start installing.
   1. Or, select more items, which will be placed as the first item, with the ` (selected)` tag.
1. Optionally, you can undo your selection by re-selecting a item with ` (selected)` tag. Then, this excluded item will be placed as the first option again, but this time with the ` (excluded)` tag.
1. When I will actually install things, I do not re-parse the whole list again. I only do it from the first item up to the selected items count, checking whether they had the ` (excluded)` tag. If so, I skip them. If not, I just remove the ` (selected)` tag and add them to the installation list.
1. Finally, after building a clean installation list, I install everything calling the `AdvancedInstallPackageThread`.

FichteFoll:
Thanks for the clarification on how `(selected)` and `(excluded)` work. I think adding the selected packages to the top of the list is a good idea. That also makes my suggestion to list the selected packages in the first item redundant.

However, I'm not sure how I feel about the then un-selected packages going there as well (and receiving the `(excluded)` special treatment). I guess I would have to test this a bit for myself until I can formulate an opinion on it.

deathaxe:
I've tried that out but found it rather distracting to need to select a package and then run installation with another command. It may be an interesting workflow when installing multiple packages, but is this the general case?

It may then just be easier to just re-open normal "Install Package" quick panel while a former selected package is being installed.

PR #1450, opened 2019-10-11
Output Package Control Messages without empty whitespace

dnicolson:
This pull request changes the way "Package Control Messages" are displayed, it avoids outputting whitespace when displaying blank newlines.

If the [Trailing Spaces](https://github.com/SublimeText/TrailingSpaces) package is installed you can see the before and after:

PixelSnap 2019-10-11 at 23 09 33@2x

### DIFF
--- a/package_control/package_manager.py
+++ b/package_control/package_manager.py
@@ -1755,7 +1755,7 @@ class PackageManager():
         def read_message(message_path):
             with open_compat(message_path, 'r') as f:
                 lines = read_compat(f).rstrip().split('\n')
-                return '\n' + '\n'.join(list(map(lambda s: '  ' + s if s else '', lines)))
+                return '\n' + '\n'.join(['  ' + s if s else '' for s in lines])
 
         output = ''
         if not is_upgrade and message_info.get('install'):
--- a/package_control/package_manager.py
+++ b/package_control/package_manager.py
@@ -1754,7 +1754,8 @@ class PackageManager():
 
         def read_message(message_path):
             with open_compat(message_path, 'r') as f:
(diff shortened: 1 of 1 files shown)


package_control/package_manager.py
```diff
@@ -1754,7 +1754,8 @@ def print_messages(self, package, package_dir, is_upgrade, old_version, new_vers
 
         def read_message(message_path):
             with open_compat(message_path, 'r') as f:
-                return '\n  %s\n' % read_compat(f).rstrip().replace('\n', '\n  ')
+                lines = read_compat(f).rstrip().split('\n')
+                return '\n' + '\n'.join(list(map(lambda s: '  ' + s if s else '', lines)))
```
FichteFoll: This is a better case for a generator comprehension than a filter call with a lambda. (Also `join` accepts an iterable.)

package_control/package_manager.py
```diff
@@ -1754,7 +1754,8 @@ def print_messages(self, package, package_dir, is_upgrade, old_version, new_vers
 
         def read_message(message_path):
             with open_compat(message_path, 'r') as f:
-                return '\n  %s\n' % read_compat(f).rstrip().replace('\n', '\n  ')
+                lines = read_compat(f).rstrip().split('\n')
+                return '\n' + '\n'.join(['  ' + s if s else '' for s in lines])
```
FichteFoll:
```suggestion
                return '\n' + '\n'.join('  ' + s if s else '' for s in lines)
```

Generator expressions are iterable as well.

dnicolson:
@wbond would you mind reviewing this change?

wbond:
Look good to me, thanks!

PR #1472, opened 2020-06-03
Fix n² directory checks during package installation

BenjaminSchaaf:
Dramatically improves install times for packages with large amounts of
files. Takes it from 1+ hours to <1 minute.

### DIFF
--- a/package_control/package_manager.py
+++ b/package_control/package_manager.py
@@ -1292,8 +1292,9 @@ class PackageManager():
 
             # Here we don't use .extractall() since it was having issues on OS X
             overwrite_failed = False
-            extracted_paths = []
-            for path in package_zip.namelist():
+            extracted_paths = set()
+            for info in package_zip.infolist():
+                path = info.filename
                 dest = path
 
                 try:
@@ -1349,7 +1350,7 @@ class PackageManager():
 
                 def add_extracted_dirs(dir_):
                     while dir_ not in extracted_paths:
-                        extracted_paths.append(dir_)
+                        extracted_paths.add(dir_)
                         dir_ = os.path.dirname(dir_)
                         if dir_ == package_dir:
                             break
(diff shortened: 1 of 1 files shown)


package_control/package_manager.py
```diff
@@ -1415,6 +1416,7 @@ def add_extracted_dirs(dir_):
             # if there is an error removing a file. The next time there is an
             # upgrade, it should be cleaned out successfully then.
             clear_directory(package_dir, extracted_paths)
+            clear_directory(package_dir, set(extracted_paths))
```
deathaxe, line 1419: What's the intention of calling `clear_directory` twice for `extracted_paths`?

BenjaminSchaaf: Looks like I just screwed up line staging :)

package_control/package_manager.py
```diff
@@ -1292,8 +1292,9 @@ def install_package(self, package_name, is_dependency=False):
 
             # Here we don't use .extractall() since it was having issues on OS X
             overwrite_failed = False
-            extracted_paths = []
-            for path in package_zip.namelist():
+            extracted_paths = set()
+            for info in package_zip.infolist():
+                path = info.filename
                 dest = path
```
FichteFoll, line 1298: *Could* be shortened to `dest = path = info.filename`, but really does't matter.

package_control/package_manager.py
```diff
@@ -1292,8 +1292,9 @@ def install_package(self, package_name, is_dependency=False):
 
             # Here we don't use .extractall() since it was having issues on OS X
             overwrite_failed = False
-            extracted_paths = []
-            for path in package_zip.namelist():
+            extracted_paths = set()
+            for info in package_zip.infolist():
```
wbond, line 1296: Is there a performance difference in the call to `infolist()` versus `namelist()`, or is it just in the usage of `set()`?

If it is important to call `infolist()`, it would be good to add a comment explaining why we use it so it isn't refactored away in the future to be more "pythonic".

BenjaminSchaaf: It simply avoids creating a whole new list from `infolist()`. It's implemented as `[info.filename for info in infolist()]`. So it's not very important, but I felt that if the zip `info` is ever going to be required in this loop it'll be done right if it's already available - rather than someone querying the info by filename for every file.

wbond:
Awesome, thanks for the fix!

PR #1480, opened 2020-07-15
[Fix] downloaders not using the configured User Agent

TheSecEng:
fix issue with downloaders not using the user_agent setting. Also falling back to a default setting if the user overrides there setting with `""`

### DIFF
--- a/package_control/download_manager.py
+++ b/package_control/download_manager.py
@@ -29,6 +29,7 @@ from .downloaders.downloader_exception import DownloaderException
 from .downloaders.win_downloader_exception import WinDownloaderException
 from .downloaders.oscrypto_downloader_exception import OscryptoDownloaderException
 from .http_cache import HttpCache
+from .settings import pc_settings_filename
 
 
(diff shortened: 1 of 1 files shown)


wbond:
Can you provide a debug log of the user agent not being set? I don't see this error on my end.

Right now the fix is injecting settings into a place it shouldn't be. The website uses a different settings source, so unequivocally loading from `pc_settings_filename()` would break that.

TheSecEng:
> Can you provide a debug log of the user agent not being set? I don't see this error on my end.
> 
> Right now the fix is injecting settings into a place it shouldn't be. The website uses a different settings source, so unequivocally loading from `pc_settings_filename()` would break that.

Sure I can upload a log today, I only noticed it when testing the Gitlab client

TheSecEng:
Can't replicate, maybe it was in the Tests that I ran into this.

TheSecEng:
@wbond I believe that the Unit Tests are where I experienced this now that I think about it.

PR #1497, opened 2020-09-03
Fix QuickPanelItem usages

deathaxe:
This commit fixes a regression with several commands which use quick panel items' labels, which were broken by commit d22abdbab4e7a6f071b485f5ec49319abb986c12.


wbond:
Thanks for taking the time to do this!

PR #1505, opened 2020-09-26
[4.0] Remove remaining unicode string literals

deathaxe:


rwols:
Why was this closed?

deathaxe:
I guess because the target branch was renamed.

deathaxe:
There are dozens of commits from `next` branch to check and re-bundle to drop py2 support from `four-point-oh`. It's only a little part of that.

PR #1521, opened 2021-02-01
Replace Travis by GitHub workflows

deathaxe:
It appears travis-ci.org is closing in some weeks.

Maybe a good point to switch to github actions for linting pull requests.


deathaxe:
flake8 fails on unused but assigned variables in the new dependency.py module.

FichteFoll:
I suspect that workflow pull request involving the secrets context must be on the `master` branch before they are active.

I've also had success with https://github.com/rbialon/flake8-annotations, just as another example.

wbond:
I appreciate you trying to fix things, but can you explain why you removed the ability for me to installing linting tools and run the linter for Python 3.3?

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,22 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
```
wbond, line 16: This should be 3.3.7, or alternatively 3.3.7 and 3.8.

dev/deps.py
```diff
@@ -1,881 +0,0 @@
-# coding: utf-8
```
wbond, line 1: This file should be restored, unless you can articulate how you've replaced all of the functionality.

dev/lint.py
```diff
@@ -1,55 +0,0 @@
-# coding: utf-8
```
wbond, line 1: This file should be restored, unless you can articulate how you've replaced all of the functionality.

dev/requires
```diff
@@ -1,18 +0,0 @@
-setuptools >= 39.0.1 ; python_version == '2.7' or python_version >= '3.3'
```
wbond, line 1: This file should be restored, unless you can articulate how you've replaced all of the functionality.

package_control/deps/oscrypto/_mac/_security_ctypes.py
```diff
@@ -16,7 +16,7 @@
 
 
 version = platform.mac_ver()[0]
-version_info = tuple(map(int,  platform.mac_ver()[0].split('.')))
```
wbond, line 19: This will need to be addressed upstream

package_control/deps/oscrypto/_pkcs5.py
```diff
@@ -127,7 +127,7 @@ def pbkdf2(hash_algorithm, password, salt, iterations, key_length):
 
         u = int_from_bytes(last)
 
-        for _ in range(iterations-1):
+        for _ in range(iterations - 1):
```
wbond, line 130: This will need to be addressed upstream

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,22 @@
...
+    - name: Install dependencies
+      run: pip install flake8 lintly
```
wbond, line 18: This won't work on Python 3.3. You should probably use the previously developed "deps.py".

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,22 @@
...
+    - name: Lint with flake8
+      run: flake8 --config=tox.ini . | lintly
```
wbond, line 20: Lintly will have to be removed since it doesn't support 3.3. That or add different commands for 3.3 and 3.8.

wbond (changes requested):

wbond:
> flake8 fails on unused but assigned variables in the new dependency.py module.

This is known. I pushed up some WIP and wanted to leave those there because there is the intention they will be used.

.github/workflows/pull_request.yml
deathaxe: You mean the dev folder? I interpreted it as scripting folder used to run travis-ci only and thus obsolete by moving to git workflows.

Linting sources locally can easily be achieved by just running `flake8 --config=tox.ini` via command line, which can be installed via `pip install flake8`. 

Is there more happening in the `dev` package, which I am not aware of?

.github/workflows/pull_request.yml
deathaxe: Those scripts run online. Is it really required them to run on 3.3? I am not aware of linting rules differ between 3.3 and 3.8.

wbond: Yes, Python 3.8 has all kinds of syntax that doesn't exist on 3.3.

wbond: If you only lint for Python 3.8, we effectively are EOLing ST3.

.github/workflows/pull_request.yml
wbond: It makes it possible for users developing it locally with a Python 3.3 install to easy get the properly packages to lint, and run the linting rules. Since Python 3.3 is unsupported by a bunch of packages, you need specific versions, and pip even has issues on various CIs I used with various Python packages I develop.

The idea is that we'll probably add AppVeyor and CircleCI to make sure tests are being run on all platforms.

deathaxe:
> This is known. I pushed up some WIP

Sure, just wanted to note it to explain failing tests.

.github/workflows/pull_request.yml
wbond: Just someone adding an F string will break 3.3.

.github/workflows/pull_request.yml
rwols: > The idea is that we'll probably add AppVeyor and CircleCI to make sure tests are being run on all platforms.

You can run windows and macOS jobs with github workflows.

deathaxe: > The idea is that we'll probably add AppVeyor and CircleCI

Which renders the whole attempt here useless, so we can close it.

.github/workflows/pull_request.yml
deathaxe: That's a good point.

deathaxe:
If all of that is too much more than handling a git workflow It is not for me.

.github/workflows/pull_request.yml
wbond: > You can run windows and macOS jobs with github workflows.

The downside being GitHub only supports very new releases, and we've definitely hit OS release-specific issues.

wbond:
No worries, I appreciate you trying to help. It sounds like lintly may be useful to add to a 3.8 run.

I'll whip up a PR since I've (unfortunately) had way too much experience with the "joy" of CI on Python projects.

deathaxe:
World needs more efficent solutions to share minds. :-D

Those topics just start to get a need for my part of software development and I try to get some useful entry to all of that.

I just got aware of python 3.3 to be needed offline, too, in order to ensure not to break something.

PR #1524, opened 2021-02-03
Add GitlabClient to RepositoryProvider

deathaxe:
This PR adds GitlabClient to the list of clients which are used to download repository information or details about the package/dependency.

It also adds related RepositoryProvider and ChannelProvider tests.

Requires: https://github.com/wbond/package_control-json/pull/1

PR targets 3.4 as the issue is treated as bug, but I am not sure whether it is still intended to maintain such things or if it is to be used for 4.0 only.


wbond:
I merged this into four-point-oh since I don't plan to do another 3.x release. ST2 support is now officially over for Package Control, and PC 4.0 will support Sublime Text 3.0, which goes back like 4 years in terms of support.

Issue #1525, opened 2021-02-07
Add support for dependency only repositories


deathaxe:
A `RepositoryProvider` fails validating a _repository.json_ which does not contain `packages`. Repositories with `dependencies` only are therefore claimed invalid.

https://github.com/wbond/package_control/blob/8b947d227bfee2b514283e650c3f88c954ae1026/package_control/providers/repository_provider.py#L221-L224

**Example**

```
{
	"schema_version": "3.0.0",
	"dependencies": [
		{
			...
		}
	],
}

deathaxe:
It appears missing `dependencies` or `packages` lists are added by `fetch()`.

https://github.com/wbond/package_control/blob/8b947d227bfee2b514283e650c3f88c954ae1026/package_control/providers/repository_provider.py#L126-L129

PR #1527, opened 2021-02-07
Update scheme to 4.0.0

### LINKED ISSUE #1516: Update Dependency Schemas
In order to get Python 3.8 support working properly, we need dependency releases to indicate what python versions they support.

A new `python_versions` key should be added to the repository JSON schema.

The current state of the schema is: https://github.com/wbond/package_control/blob/four-point-oh/example-repository.json#L371-L427.

The code that consumes the JSON is at: https://github.com/wbond/package_control/blob/four-point-oh/package_control/providers/repository_provider.py.

The channel JSON schema also needs to be updated, since it caches the repository schema. The example is at: https://github.com/wbond/package_control/blob/four-point-oh/example-channel.json#L43-L67.

The code to consume it is at: https://github.com/wbond/package_control/blob/four-point-oh/package_control/providers/channel_provider.py.

deathaxe:
Fixes #1516 

This PR bumps repository and channel scheme version to 4.0.0

1. Scheme versions are validated using `SemVer` class to be able to access major, minor, patch directly.
2. Json keys and related methods are renamed to use `libraries` instead of `dependencies`.
3. Library releases must provide `python_versions: ['3.3', ...]` as of scheme version `4.0.0`.
4. Libraries must not provide `load_order` as of scheme version `4.0.0`.
5. Included repositories' scheme version is validated and must match the scheme version of the including repository. That's required as several scheme changes are handled by checking the root repository's scheme version only.

Notes:

- [x] New integration tests for scheme 4.0.0 need https://github.com/wbond/package_control-json/pull/2
- [x] https://github.com/wbond/package_control/pull/1520 is needed to run integration tests successfully.
- [x] Not directly related but https://github.com/wbond/package_control/pull/1524 may need to be taken into account as well.

### DIFF
--- a/package_control/providers/repository_provider.py
+++ b/package_control/providers/repository_provider.py
@@ -24,7 +24,7 @@ class InvalidRepoFileException(ProviderException):
     def __init__(self, repo, reason_message):
         super().__init__(
             'Repository %s does not appear to be a valid repository file because'
-            ' %s' % (repo.repo, reason_message))
+            ' %s' % (repo.repo_url, reason_message))
 
 
 class RepositoryProvider(BaseRepositoryProvider):
--- a/package_control/providers/repository_provider.py
+++ b/package_control/providers/repository_provider.py
@@ -328,12 +328,14 @@ class RepositoryProvider():
         output = {}
         for dependency in self.repo_info['dependencies']:
             info = {
+                'releases': [],
                 'sources': [self.repo]
             }
 
-            for field in ['name', 'description', 'author', 'issues', 'load_order']:
-                if dependency.get(field):
-                    info[field] = dependency.get(field)
+            for field in ('name', 'description', 'author', 'issues', 'load_order'):
+                field_value = dependency.get(field)
+                if field_value:
+                    info[field] = field_value
 
             if 'name' not in info:
                 self.failed_sources[self.repo] = ProviderException(text.format(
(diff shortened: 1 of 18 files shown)


package_control/providers/repository_provider.py
```diff
@@ -168,60 +167,27 @@ def fetch_and_validate(self):
...
+            except KeyError:
+                fail('the "schema_version" JSON key is missing.')
+            except ValueError as e:
+                fail(e)
 
-        if 'packages' not in self.repo_info:
-            error_string = '%s the "packages" JSON key is missing.' % schema_error
-            fail(error_string)
-            return False
+            if isinstance(self.repo_info['packages'], dict):
```
FichteFoll, line 184: What happened to the `if 'packages' not in self.repo_info:` check? If you meant to make the key optional, you must consider that case before accessing `self.repo_info['packages']` or use `dict.get`.

FichteFoll (commented):
Personally I'm not a fan of huge try-except blocks, but the previous code already has very long functions and try-except does seem like the best option, even without refactoring out a couple blocks into functions.

package_control/providers/repository_provider.py
deathaxe: Please read commit messages to learn about the answer (hint: aa05f44). It also reveals #1525 to be a result of incomplete reading and therefore a false positive.

deathaxe:
The methods in general are quite large and should probably be splitted in order to get some testable pieces, but I don't want to refactor everything just for nothing at this point.

The try blocks are not that large, IMHO. It is exactly the use case of try...catch to wrap algorithms of one or limited purpose which needs to finish en block or fail early without the need to write deeply nested if..else or goto statements. The only reason for the size of the block is the number of objects handled explicitly. All operations are of the same purpose - validating content. So I don't see any issues. IMHO, it's even more consistent this way than trying to `continue` nested loops and wait to abort execution until the loop is done.

The try catch block could easily be reduced to one line by moving the code to another method, if it is a visual problem.

FichteFoll:
The PR is fine to me as-is. I mostly asked a question out of curiosity. Of course, splitting would be nice, but not necessarily within the scope of this PR.

wbond:
I just rebased this onto latest `four-point-oh` to get the CI working. Hoping to review in the next few days.

Thanks for all of the improvements and clean you've done in these PRs @deathaxe!!

wbond:
Based on feedback related to mdpopups and people putting package resources in dependencies, I propose:

 1. We use schema version 4.0.0
 2. Current dependencies will be renamed to “libraries” and load_order will be dropped.

This will open up the term “dependency” for someone to implement inter-package dependencies. That will benefit packages that want to do things like mdpopups, and make installing things like LSP and SublimeLinter easier. I’m not proposing we need to implement the new dependencies now, just prepare the schema.

deathaxe:
This also means to rename `dependencies.json` and all `dependencies` keys in package metadata and dependency related methods, while keeping backward compatibility with older schemes which still use the old term? Sounds quite invasive change.

wbond:
Yeah, it does. I’m not expecting you to do all of that, though. More interested in discussion.

Either way all of these dependency changes are going to be pretty invasive and require a bunch of beta testing.

deathaxe:
1. Basically the term `libraries` seems more precise with regards how they are meant to be used in ST. Python doesn't distinguish packages and libraries though.

2. I don't really see a use case for re-using `dependencies` in the future. 

   1. A today's dependency is a pure python package which will be installed into _Lib/_.
   2. Packages may depend on each other in the future, so backend-service packages may be pulled automatically when installing others which need them. If those shouldn't be installed manually just a flag might be needed to hide them from quick panels, that's it. I don't see a reason for another kind of package, to be honest.

3. It's all quite much effort just because of two dependency packages abusing the intended structure. I know @facelessuser is aware of required changes to keep _mdpopups_ working. The only resources `lsp_utils` provides is a default settings file with a single item and the related command to open the file. Complains about loosing that are exagurated, IMHO. Those settings should be part of the LSP-xxx helper packages as if `lsp_utils` wouldn't exist.

4. `RepositoryProvider` will need some changes as scheme version needs to be validated before reading anything from it. Also found included repositories not beeing validated at all, atm. They are just fetched and merged to `repo_info`. We should probably make sure included repositories use the same scheme version to prevent bad things from happening. Other than that I'd try to translate the `dependencies` key from older schemes into `libraries` in `fetch()` to avoid too much code branches.

facelessuser:
Just to note, mdpopups will most likely move to vendoring what it needs in the future. It's just kind of a pain managing that chain, at least for that package. I'd like it to be less sensitive to the libraries updating that it depends on as well. While many of the things it depends on are useful outside of mdpopups, and fine to be standalone dependencies/libraries in ST, I find being able to lock mdpopups to specific, vendored versions will be easier to manage.

deathaxe:
Ok, renamed a couple of things to move towards _libraries_. Installing/removing packages/deps still seems to work. Can't run integration tests atm. due to server errors.

wbond:
I've:

 - rebased this onto recent fixes
 - fixed tests
 - merged and updated https://github.com/wbond/package_control-json/pull/2

wbond:
I should note that I did a force push

PR #1528, opened 2021-02-07
Update .gitignore and add .gitattributes

deathaxe:
Some packages or tools create cache files/folders within the project which just can be ignored by git.

Other files are crucial for the repository but don't need to be part of Package Control.sublime-package. They are added to the .gitattributes file to not export them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ dev/.lint-deps/
 
 # linter cache
 .mypy_cache/
+.pytest_cache/
 .ropeproject/
 
 # python cache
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,6 +1,16 @@
...
+__pycache__/
 *.pyc
+*.pyo
+
+# Sublime Text files
 *.sublime-project
-*.thTheme.cache
 *.sublime-workspace
-*.mypy_cache/
+*.thTheme.cache
```
rwols, line 16: tmTheme.cache?

deathaxe: Existed there before. Hence kept it.

PR #1531, opened 2021-02-07
Remove 'open_compat' module

deathaxe:
This PR...

1. removes `open_compat` and `file_not_found_error` module which were required for compatibility with python 2 only.
2. use `fobj` as file object variable instead of `f` or arbitrary other characters.
3. pass encoding to `open(..., encoding='utf-8')` instead of manually handling that
4. Prefer `json.load(fobj)` over `json.loads(fobj.read())`
5. Create empty files in binary mode. No need to allocate an encoder for writing nothing.


package_control/ca_certs.py
```diff
@@ -42,19 +41,19 @@ def get_ca_bundle_path(settings):
         regenerate = regenerate or os.path.getmtime(user_ca_bundle_path) > os.path.getmtime(merged_ca_bundle_path)
 
     if regenerate:
-        with open(merged_ca_bundle_path, 'wb') as merged:
+        with open(merged_ca_bundle_path, 'w', encoding='utf-8') as merged:
             if system_ca_bundle_path:
-                with open_compat(system_ca_bundle_path, 'r') as system:
-                    system_certs = read_compat(system).strip()
-                    merged.write(system_certs.encode('utf-8'))
+                with open(system_ca_bundle_path, 'r', encoding='utf-8', errors='replace') as system:
```
FichteFoll, line 46: I have no clue about the ca bundles code, but do we really want `errors='replace'` here and below?

package_control/dependency.py
```diff
@@ -101,13 +101,13 @@ def install(dest_root, src_dir, name, version, description, url, plat_specific):
 
     di_record_path = os.path.join(dist_info_path, 'RECORD')
     # Create an empty file so it shows up in its own file list
-    with open(di_record_path, 'wb') as f:
-        f.write(b'')
+    with open(di_record_path, 'w', encoding='utf-8') as fobj:
```
FichteFoll, line 104: This could have remained as using binary mode. In fact, the following line can probably be removed/replaced with `pass` as well.

package_control/downloaders/wget_downloader.py
```diff
@@ -235,8 +234,8 @@ def parse_output(self, clean_run):
             HTTP header names.
         """
 
-        with open_compat(self.tmp_file, 'r') as f:
-            output = read_compat(f).splitlines()
+        with open(self.tmp_file, 'r', encoding='utf-8', errors='replace') as fobj:
+            output = fobj.read().splitlines()
```
FichteFoll, line 238: `fobj.readlines()`? Not sure off-hand whether that keeps line endings.

package_control/ca_certs.py
deathaxe: That's what `open_compat` used to do and I didn't intend to change behavior.

I can barely remember of an issue with regards to certain unicode characters contained in certificates but I am not aware of whether using `replace` was a fix for it. Basically I wonder, whether those could simply be copied in binary mode, but I guess, it would make stripping harder.

package_control/dependency.py
deathaxe: Sure. I am not even sure, whether `write` is even required with regards to other usages which just do `open().close()` with same result.

package_control/downloaders/wget_downloader.py
deathaxe: Yes, it does.

package_control/ca_certs.py
deathaxe: This behavior of `open_compat()` exists since day 1 (9432d8184a1b64c4b541ca1bd18ffda9e63e2875) of ST3 support.

package_control/dependency.py
FichteFoll: Seeing how often empty files are created, a utility function to do just that seems reasonable, but of course not required.

deathaxe: Not sure whether it's really worth it.

dev/deps.py
```diff
@@ -292,8 +292,8 @@ def _extract_package(deps_dir, pkg_path, pkg_dir):
                     dst_dir = os.path.dirname(dst_path)
                     if not os.path.exists(dst_dir):
                         os.makedirs(dst_dir)
-                    with open(dst_path, 'wb') as f:
```
wbond, line 295: `deps.py` is actually shared across a number of different Python projects I write, some of which still support Python 2, so I'd prefer to leave it alone so I can more easily migrate fixes in the future.

package_control/ca_certs.py
wbond: We can kill `errors='replace'`. I'd rather get an exception and know about it.

package_control/ca_certs.py
```diff
@@ -42,19 +41,19 @@ def get_ca_bundle_path(settings):
...
+                    system_certs = system.read().strip()
+                    merged.write(system_certs)
                     if len(system_certs) > 0:
-                        merged.write(b'\n')
+                        merged.write('\n')
             if os.path.exists(user_ca_bundle_path):
-                with open_compat(user_ca_bundle_path, 'r') as user:
-                    user_certs = read_compat(user).strip()
-                    merged.write(user_certs.encode('utf-8'))
+                with open(user_ca_bundle_path, 'r', encoding='utf-8', errors='replace') as user:
```
wbond, line 52: Similar here, remove the `errors` arg.

package_control/downloaders/curl_downloader.py
```diff
@@ -142,8 +141,8 @@ def download(self, url, error_message, timeout, tries, prefer_cached=False):
             try:
                 output = self.execute(command)
 
-                with open_compat(self.tmp_file, 'r') as f:
-                    headers_str = read_compat(f)
+                with open(self.tmp_file, 'r', encoding='utf-8', errors='replace') as fobj:
```
wbond, line 144: I looked and could not find any reference to curl explicitly encoding headers as UTF-8. HTTP should be using ISO-8859-1, so we should use that when reading from the headers.

package_control/downloaders/wget_downloader.py
```diff
@@ -235,8 +234,8 @@ def parse_output(self, clean_run):
             HTTP header names.
         """
 
-        with open_compat(self.tmp_file, 'r') as f:
-            output = read_compat(f).splitlines()
+        with open(self.tmp_file, 'r', encoding='utf-8', errors='replace') as fobj:
```
wbond, line 237: Similar to curl, I looked into wget and couldn't find anything about the encoding of the log file. Since we are grabbing HTTP headers, we should probably decode as ISO-8859-1.

wbond (changes requested):
Overall looks good.

There are two places we should probably change the encoding.

package_control/downloaders/curl_downloader.py
deathaxe: utf-8 is what read_compat used to use. Changing to ISO-8859-1 shouldn't make an difference in case headers use ascii charset only.

How about reading from stderr then? It appears everything is decoded as utf-8 explicitly. May it be safe to just omit `encoding` argument to let python choose the best encoding for each OS?

deathaxe: ... see line 165 for instance.

package_control/downloaders/wget_downloader.py
deathaxe: Removed `errors='replace'` here as well. Hope that's ok?

wbond: Yeah, honestly these are probably very rarely used these days now that ST 3.0+ has the SSL module built in on Linux

wbond:
Looks good to me, thanks for cleaning this up!

PR #1538, opened 2021-02-23
Fix http cache path creation

deathaxe:
Fixes https://github.com/sublimehq/sublime_text/issues/3964

This PR fixes an error which causes http_cache folder not to be created in case the whole Cache/Package Control path was removed during runtime.


wbond:
Thanks!

PR #1539, opened 2021-02-23
Fix cli_downloader on Windows

deathaxe:
This PR enables usage of cli_downloader on Windows OS by fixing 2 issues.

First of them prevented `curl.exe` or `wget.exe` from being found by `find_binary()` method as it missed to add the file extension.

Second fix addresses an issue with console window being displayed if an executable is invoked, which caused bad user experience.

curl.exe is shipped by default as of Windows 10 and therefore might be used as downloader. May be quite uncommon as the others do a good job, but if someone decides to use it, it should work.

### DIFF
--- a/package_control/downloaders/cli_downloader.py
+++ b/package_control/downloaders/cli_downloader.py
@@ -43,13 +43,16 @@ class CliDownloader(object):
             # This is mostly for OS X, which seems to launch ST with a
             # minimal set of environmental variables
             dirs.append('/usr/local/bin')
+            executable = name
+        else:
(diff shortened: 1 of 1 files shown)


rwols:
These downloaders are core to Package Control, can't we build some integration tests around these things?

deathaxe:
You want something like that: https://github.com/deathaxe/package_control/blob/next/package_control/tests/downloaders.py?

wbond:
I don't really have the intention of trying to support curl or wget on Windows, but if someone wants to configure them, I am fine with that.

PR #1545, opened 2021-02-26
Fix ST3 compatibility issue of bootstrapper

deathaxe:
ST3 uses `Libs/python3.3` while ST4 uses `Libs/python33`, so destination for package control injector needs to be tweaked a little bit depending which ST version is used.

### DIFF
--- a/2_bootstrap.py
+++ b/2_bootstrap.py
@@ -209,21 +209,21 @@ def _install_injectors(settings):
     """
 
     injector_code = dedent(injector_code).strip() + "\n"
+    injector_code = injector_code.encode('utf-8')
 
-    injector_path_33 = os.path.join(sys_path.data_dir, 'Lib', 'python33', 'package_control.py')
-    try:
-        if not os.path.exists(injector_path_33):
-            with open(injector_path_33, 'wb') as f:
-                f.write(injector_code.encode('utf-8'))
-    except (OSError) as e:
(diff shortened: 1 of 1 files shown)


wbond:
This sounds like a backwards compatibility break in core

deathaxe:
Yes, it is. I can barely remember someone talking about certifying for MacOS to be a reason for the naming change, but I am not sure. The one way or the other, the folder names of `<bindir>/Data/Libs/python...` and `<bindir>/Libs/python/...` should stay equal, IMHO - in case you plan to rename those files in ST4. ;-)

FichteFoll:
I think I mentioned this after the first few builds of ST4 with the two python hosts on Discord, but can't find it, nor remember whether that was discussed afterwards. Since the period version appears to be the standard for Python distribitions on Linux at least (i.e. `/usr/lib/python3.8` etc) and that's also what ST3 had, ST4 should probably revert to that before long.

wbond:
I believe @deathaxe is correct. We can’t use periods in the MacOS folder and notarize. We didn’t run into that with ST3 because there were no file paths with the period. Since PC is the only thing that is affected by this and it is a clean break at build 4050, I think we are good.

deathaxe:
A common naming scheme for python files is what matters, imho. On Windows we have python33.dll and python38.dll so naming the library paths the same way is a logical step. Only the zipped stdlib archives don't fit well into that as those use periods in the name, atm.

PR #1551, opened 2021-03-07
Fix migration error when a src dir already exists

### LINKED ISSUE #1549: Exception when migrating deps in four-point-oh branch
The following exception occurs on startup.

```
Package Control: Error trying to migrate dependencies - [Errno 17] File exists: '/Users/raoulwols/Library/Application Support/Sublime Text 3/Lib/python33/yaml'
Exception in thread Thread-3:
  File "./python3.3/threading.py", line 901, in _bootstrap_inner
  File "./python3.3/threading.py", line 858, in run
  File "/Users/raoulwols/Library/Application Support/Sublime Text 3/Packages/Package Control/2_bootstrap.py", line 113, in _migrate_loaders
    plat_specific
  File "/Users/raoulwols/Library/Application Support/Sublime Text 3/Packages/Package Control/package_control/dependency.py", line 94, in install
    shutil.copytree(source, os.path.join(dest_root, dname))
  File "./python3.3/shutil.py", line 301, in copytree
  File "./python3.3/os.py", line 262, in makedirs
FileExistsError: [Errno 17] File exists: '/Users/raoulwols/Library/Application Support/Sublime Text 3/Lib/python33/yaml'
```

rwols:
close #1549

### DIFF
--- a/package_control/dependency.py
+++ b/package_control/dependency.py
@@ -91,7 +91,9 @@ def install(dest_root, src_dir, name, version, description, url, plat_specific):
     package_dir_names = []
     for dname, source in package_dirs:
         package_dir_names.append(dname)
-        shutil.copytree(source, os.path.join(dest_root, dname))
+        destination = os.path.join(dest_root, dname)
+        shutil.rmtree(destination, ignore_errors=True)
+        shutil.copytree(source, destination)
 
     package_file_names = []
     for rel_dest, source in package_files:


deathaxe:
I guess this issue will be resolved once new dependencies are installed into libs directly. Fixing it at this point might cause all deps to be migrated at each startup, which is probably not what we want.

rwols:
> Fixing it at this point might cause all deps to be migrated at each startup, which is probably not what we want.

That's what I was wondering about. Why is PC still running the migrate code once `"bootstrapped"` is set to `4` in Packages/User/Package Control.sublime-settings?

deathaxe:
Because it just checks for existence of `0_package_control_loader` package, which is re-created as soon as dependencies are updated. You can't expect perfect behavior at this point. It's a WIP branch. 😄

rwols:
OK, it sounds like there was some thought put it this at least.

PR #1554, opened 2021-03-11
PackageManager Cleanups

deathaxe:
1. Some methods use `package` some `package_name` as argument names. This PR proposes to use `package_name` for all of them to make clear a name string is meant.
2. removes some `package_file_exists()` calls, which are already handled by the gated methods and thus reduntant.
3. drops another ST2 construct which pushed adding/removing `installed_packages` to main thread via `set_timeout()`.

### DIFF
--- a/package_control/package_manager.py
+++ b/package_control/package_manager.py
@@ -1404,7 +1404,7 @@ class PackageManager():
                 # settings across computers and have the same packages installed
                 settings = sublime.load_settings(pc_settings_filename())
                 names = load_list_setting(settings, 'installed_packages')
-                if package_name in names:
+                if package_name not in names:
                     names.append(package_name)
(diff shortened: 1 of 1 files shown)


rchl:
> 3\. drops another ST2 construct which pushed adding/removing `installed_packages` to main thread via `set_timeout()`.

I know almost nothing about Package Control but are you sure that's only needed for ST2 and not some fix for some subtle timing issue?

deathaxe:
Yes. It wasn't possible to call API functions from threads in ST2. That's why all those were delegated to main thread this way. I guess that's also the main reason for caching all package control settings in `__init__()`.

deathaxe:
Rebased that branch, too.

wbond:
These changes look good, thank you!

PR #1559, opened 2021-04-22
fixing Gitlab unit tests and misconfigurations

TheSecEng:
Fixed some Gitlab tests that had been failing, also implemented the GitlabClientTests which was implemented but not setup to run.


TheSecEng:
Fixed in four-dot-oh

PR #1561, opened 2021-04-27
Update dependencies on doing the update check

rchl:
The dependencies were not updated on running the update check. They were
only updated if some package that depended on them was being updated but
that is not sufficient.

Note that I haven't tested this change on `four-point-oh` branch because that branch is not functional at the current time. I've tested on `master` though and the relevant code looks the same and the fix applies cleanly to both so I'm assuming it should work as well.

As this appears to be quite a high-level code, it possibly won't even be affected (broken) during a rewrite of the dependency handling.


package_control/automatic_upgrader.py
```diff
@@ -332,3 +333,30 @@ def disable_packages():
...
+                'downgrade',
+                'overwrite',
+                'none'
+            ],
+            ignore_packages=self.auto_upgrade_ignore,
+            get_dependencies=True
+        )
+
+        if USE_QUICK_PANEL_ITEM:
+            dependency_list = [info.trigger for info in dependency_list]
```
rchl, line 358: Note: Here and in the original `upgrade_packages` method, the handling of `USE_QUICK_PANEL_ITEM` seems pointless because this is the automatic upgrade and it never presents this info in the UI. So handling of it here and in `PackageInstaller.make_package_list` could be removed to simplify things.

FichteFoll: Isn't it just to be able to interpret the result of `self.installer.make_package_list`, which returns a `List[QuickPanelItem]` if this flag is true?

package_control/automatic_upgrader.py
```diff
@@ -332,3 +333,30 @@ def disable_packages():
                     ''',
                     (package_name, version)
                 )
+
+    def upgrade_depenendencies(self):
```
FichteFoll, line 337: "dependencies" has been renamed to "libraries"

package_control/package_installer.py
```diff
@@ -50,15 +50,23 @@ def make_package_list(self, ignore_actions=[], override_action=None, ignore_pack
...
               0 - package name
               1 - package description
               2 - action; [extra info;] package url
         """
 
-        packages = self.manager.list_available_packages()
-        installed_packages = self.manager.list_packages()
+        if get_dependencies:
+            packages = self.manager.list_available_dependencies()
+            installed_packages = self.manager.list_dependencies()
```
FichteFoll, line 66: These functions have been renamed on the four-point-oh branch.

https://github.com/wbond/package_control/blob/79c6275f1fdc28ad3aa9d1939a6d94dedc9c6c20/package_control/package_manager.py#L175

FichteFoll (dismissed):

FichteFoll:
I can't really comment on the overall logic here because I'm totally not into the code base.

package_control/automatic_upgrader.py
rchl: I've mentioned that the handling could be removed in **both places** because the result of `make_package_list` is never presented in the UI so it doesn't have to return results in "quick panel" format.

package_control/package_installer.py
rchl: I did some renaming to my changes but no, those are still not renamed on four-point-oh branch.

FichteFoll: You're right, I got the wrong function there.

rchl:
@FichteFoll I suppose there are no really any changes you require me to do so you could change the approval state to not make it seem like there is more for me to do here.

rchl:
I could fix it if it matters. Or someone else can do before merging.
Not sure if I should care or not because over 1 year without any indication whether it's wanted or not is not very reassuring.

rchl:
Is this still relevant for the 4.0 branch or should I just close it?

deathaxe:
Installing or updating a package triggers install/update of all used dependencies/libraries. So this PR is probably no longer required.

rchl:
That's not the case that was being fixed here. Check my initial message. The case that was being fixed was when a new version of dependency was released, without new version of package using it being released.

deathaxe:
Libraries are not yet updated independently, but your implementation wouldn't work at all anymore as they are no longer handled as a kind of package.

I guess I kept this open as reminder. An issue would be enough though.

deathaxe:
6da06bc2ff0efe2f6e3b9b3bb5387951353e411c adds support to automatically upgrade libraries/dependencies.

rchl:
Are you sure that code will update outdated dependencies? Based on the naming of called methods I would think it will only install missing dependencies. So if that's not the case then maybe the method names are not ideal.

deathaxe:
required - missing = installed and maybe outdated

PR #1562, opened 2021-05-01
Fix unix paths in wheel RECORD

deathaxe:
With a look at python for Windows, the RECORD file stores path in unix format even on Windows. This PR fixes RECORD file creation to comply with python's default behavior.


wbond:
Looks good, thanks!

PR #1563, opened 2021-05-01
DistInfo class

deathaxe:
Does it make sense to think of the whole `.dist-info` handling functionality as a dedicated class, so that any functionality needed to read/write library related meta data can be handled by an object later on?

I mean the wheel format has several disadvantages compared to current dependencies, which make handling those more complex. It might therefore be useful to put some of those low-level functions into a class to hide them away and make handling them easier again.

It's just an idea.


wbond:
I rebased and force-pushed to fix conflicts from #1527

package_control/distinfo.py
```diff
@@ -0,0 +1,381 @@
...
+    for dir_name in os.listdir(install_root):
+        if dir_name.endswith('.dist-info'):
+            yield DistInfoDir(install_root, dir_name)
+    return False
+
+
+class DistInfoDir:
+
+    def __init__(self, install_root, dist_info_dir):
+        self.install_root = install_root
```
wbond, line 61: It would probably good to add a docstring to explain what these parameters should be

wbond (changes requested):
I think this looks fine to me, and is probably a bit more well-contained than the old structure.

There was just on place I thought some info would be useful for future codebase readers.

package_control/library.py
```diff
@@ -101,14 +84,28 @@ def install(dest_root, src_dir, name, version, description, url, plat_specific):
...
+    try:
+        record = dist_info.read_record()
+    except FileNotFoundError:
+        raise FileNotFoundError('Library {} not installed!'.format(name))
+
+    for file_name, file_hash, file_size in record:
+        abs_path = os.path.normpath(os.path.join(install_root, file_name))
+        dir_names.add(os.path.dirname(abs_path))
+        print("removing", abs_path)
+        # os.remove(abs_path)
```
FichteFoll, line 104: You probably meant to restore this. A debug log event might be useful regardless.

package_control/library.py
```diff
@@ -101,14 +84,28 @@ def install(dest_root, src_dir, name, version, description, url, plat_specific):
...
+
+    def sort_key(a):
+        return len(a.split(os.sep))
+
+    for dir_name in sorted(dir_names, key=sort_key, reverse=True):
+        print("removing", dir_name)
+        # os.remove(dirname)
```
FichteFoll, line 111: Same here.

FichteFoll (commented):
I didn't review it too closely, but other than the two comments LGTM.

package_control/library.py
deathaxe: The `remove` part is still work in progress and it is intented to keep it dry-running at this point.

PR #1574, opened 2021-05-27
fix: GitLab Repository Provider missing from the respository_provider.py

TheSecEng:
This should fix the gitlab repository provider implementation by implementing the GitLabClient into the actual repository_providers functions.


deathaxe:
This is already fixed in four-point-oh

TheSecEng:
@deathaxe Yes, I mentioned that inside of the `package_control_channel` however @wbond had requested an issue be created so that this can be resolved now while the `four-point-oh` branch is being finished.

deathaxe:
Oh, finally PC3 receives updates again?

wbond:
Thanks for this!

PR #1579, opened 2021-07-04
Fix Version Selector

### LINKED ISSUE #1578: SemVer version selection error
It appears version selection algorithm gets confused with first two release patterns if running on a ST built which exactly matches the edge case. I'd read second rule as "use `3176-` tag for all ST builts from 3176 to 4106. But it actually seems to be `3177 to 4106` atm.

Package Control 3.4.1

#### Configured releases

```
{
	"sublime_text": "<3176",
	"tags": true
},
{
	"sublime_text": "3176 - 4106",
	"tags": "3176-"
},
{
	"sublime_text": ">=4107",
	"tags": "4107-"
}
```

#### ST 3143 / ST 3177+

Displays semver correctly.

grafik

#### ST 3176

Shows timestamp instead of semver version.

grafik

deathaxe:
Fixes #1578

A range selector is an greater than or equal and smaller than or equal operation, which includes provided min/max boundaries into valid range.

It turned out `is_compatible_version()` handles `>=` vs. `>` and `<=` vs. `<` wrong in general.

Note: Not sure if this is to be handled as hofix for current release or whether should be merged to four-point-oh only.

### DIFF
--- a/package_control/commands/package_control_tests_command.py
+++ b/package_control/commands/package_control_tests_command.py
@@ -11,6 +11,7 @@ from ..tests.providers import (
     GitLabRepositoryProviderTests,
     GitLabUserProviderTests,
     RepositoryProviderTests,
+    VersionSelectorTests,
 )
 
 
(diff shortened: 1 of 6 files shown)


deathaxe:
I'd vote to add this fix to a 3.5 as well.

PR #1580, opened 2021-07-14
fix: HTML decoding error in QuickPanelItem

jfcherng:
There are some special chars in the URL.

## Reproduction Steps

1. Open command palette
2. Run `Package Control: Install Package`
3. Type `autos` for filtering
4. Check your console

image

## Env

PC: `3.4.1`
ST: 4113 Win64

### DIFF
--- a/package_control/commands/existing_packages_command.py
+++ b/package_control/commands/existing_packages_command.py
@@ -62,7 +62,10 @@ class ExistingPackagesCommand():
                 description = '<em>%s</em>' % sublime.html_format_command(description)
                 final_line = '<em>' + action + installed_version + '</em>'
                 if url_display:
-                    final_line += '; <a href="%s">%s</a>' % (url, url_display)
+                    final_line += '; <a href="%s">%s</a>' % (
+                        sublime.html_format_command(url),
+                        sublime.html_format_command(url_display),
+                    )
                 package_entry = sublime.QuickPanelItem(package, [description, final_line])
             else:
                 final_line = action + installed_version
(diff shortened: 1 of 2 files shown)


deathaxe:
```
view.show_popup("<a href=\"http://www.sublimetext.com/forum/viewtopic.php?f=5&t=2607\">test</a>")
```

... results in same output. It's the equal sign in `t=2607` which ST's minihtml API doesn't like.

Looks more like a core bug, then something Package Control could work around.

package_control/commands/existing_packages_command.py
```diff
@@ -62,7 +62,10 @@ def make_package_list(self, action=''):
                 description = '<em>%s</em>' % sublime.html_format_command(description)
                 final_line = '<em>' + action + installed_version + '</em>'
                 if url_display:
-                    final_line += '; <a href="%s">%s</a>' % (url, url_display)
+                    final_line += '; <a href="%s">%s</a>' % (
+                        sublime.html_format_command(url),
+                        sublime.html_format_command(url_display),
```
rchl, line 67: I don't think using `html_format_command` is appropriate since it's not a command that we are formatting. We just want to encode chars as entities.

Looking at the implementation of `html_format_command`, it should be enough to use `html.escape` instead.

```py
def html_format_command(cmd, args=None):
    return html.escape(format_command(cmd, args), quote=True)
```

(I guess it's not really specific to your change as I see that it was used above already but IMO it's not really necessary or appropriate.)

deathaxe:
Added a relevant fix to four-point-oh. Closing this as superseeded as PC3.x won't receive any fixes anymore.

PR #1615, opened 2022-01-21
refactoring code with For Else

idiomaticrefactoring:
refactoring code with For Else which is more pythonic, concise and efficient; how do you think this change which has practical value？


deathaxe:
This proposal has been applied to development code for next release.

PR #1622, opened 2022-08-15
Fix client side basic auth

deathaxe:
This PR fixes two bugs related with basic authentication:

1. The `http_basic_auth` setting is not cached by PackageManager and therefore not passed to downloaders.
2. Basic authentication doesn't work for WinINetDownloader

As a result users with numerous custom repositories might quickly hit rather low API rate limits.


deathaxe:
Already part of PC4.0

Issue #1348, opened 2018-03-13
Inconsistent permissions for installed package


twolfson:
In https://github.com/wbond/sublime_terminal/issues/171#issuecomment-372095756, OS X users were reporting that some of our scripts which have the executable flag set were not receiving said flag when being installed:

https://github.com/wbond/sublime_terminal/issues/171#issuecomment-372095756

Unfortunately, I don't have time at the moment to fully look into this but it sounds like something that would be relatively easy to reproduce if you have a Package Control channel already set up

The reporters are also saying that the permissions can change on reinstall

Any idea what might be causing this?

chenzeZzz:
![image](https://user-images.githubusercontent.com/24696566/45858361-9f933d80-bd8f-11e8-95b6-52a97f9d1883.png)

deathaxe:
Fixed by PC 4.0.0-beta9

PR #1654, opened 2023-10-10
[3] Set executable permissions when extracting packages

BenjaminSchaaf:
Packages with .no-sublime-package may contain native executables or shell scripts, but python's ZipFile implementation doesn't understand executable permissions.

Fixes https://github.com/wbond/package_control/issues/1348

### DIFF
--- a/package_control/package_manager.py
+++ b/package_control/package_manager.py
@@ -9,6 +9,7 @@ import datetime
 import tempfile
 # To prevent import errors in thread with datetime
 import locale  # noqa
+from stat import S_IXUSR
 
 try:
     # Python 3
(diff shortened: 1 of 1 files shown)


package_control/package_manager.py
```diff
@@ -1393,6 +1397,11 @@ def add_extracted_dirs(dir_):
                             ''',
                             package_name
                         )
+                    finally:
```
FichteFoll, line 1400: I would expect this to not happen after an exception was thrown.
```suggestion
                    else:
```

FichteFoll (approved):

deathaxe:
Closing as obsolete. Package Control 4.0 is generally released.

PR #1660, opened 2023-12-05
Fix ignored_packages setting getting changed at startup if default packages have been extracted

BenjaminSchaaf:
If the default packages aren't in a `.sublime-package` file then on startup the `"ignored_packages"` setting would always be set to `[]`.

Primarily this fixes our development environment where we don't use any `.sublime-package` files, but presumably there are also people who've extracted these folders (as unwise as that may be).


deathaxe:
Oups. Read it as an issue, not a PR.

PR #1664, opened 2024-02-02
#1663 Fix python 3.8 dependencies being overwritten by 3.3

### LINKED ISSUE #1663: How to create a dependency for python 3.8?
I have a package https://github.com/evandrocoan/RememberCommandPaletteInput which defined the .python-version file. When package control 4 try to install it, it says "depname is not available for Python 3.8".  If I remove my .python-version file, the package dependencies are installed normally.

The documentation at https://packagecontrol.io/docs/dependencies does not mention the need of .python-version for dependencies. Does my dependencies need to have the .python-version file also?

evandrocoan:
Fix #1663

When a dependency sets `'python_versions'`, it is overwritten by 3.3 only.

### DIFF
--- a/package_control/providers/channel_provider.py
+++ b/package_control/providers/channel_provider.py
@@ -403,7 +403,7 @@ class ChannelProvider:
             for library in chain(*libraries_cache.values()):
                 del library['load_order']
                 for release in library['releases']:
-                    release['python_versions'] = ['3.3']
+                    release['python_versions'] = release.get('python_versions', ['3.3'])
                 library['releases'] = version_sort(library['releases'], 'platforms', reverse=True)
 
         else:


deathaxe:
The changed code path is for decoding repositories of schema version 1.2 to 3.0.0, which do not support `python_versions` keys. It rather migrates those to schema 4.0.0 by adding a default value.

evandrocoan:
Is there a specification for schema 4.0.0 somewhere?

evandrocoan:
Thanks for the explanation on the other issue. https://github.com/wbond/package_control/issues/1663#issuecomment-1924446083

I see this looks like an example of channel 4.0 schema you provided: https://github.com/packagecontrol/channel/blob/02bc2b4cd1f9739f203b9e3003fefca39e8d2a9c/repository.json

That example should be enough for me to upgrade my schema generation. I am just curious if there is a schema specification or if I should base myself on your schema 4.0 example?

deathaxe:
Package Control ships json schemas for schemes, to provide completions via LSP an LSP-json as well as linting/validation.

A bare bone repository or channel can be created via command palette.

There are no plans for detailed specification papers.

PR #1694, opened 2025-04-06
Teach `http_get` to read "file:///" urls

kaste:
We can already host local repositories -- i.e. "repositories" in the PC settings understand absolute local paths as additional repositories. But then inside such a repository, we cannot point a release url to a local file but MUST use the http-protocol.

E.g. for testing purposes, that means to create a local file server just to host such release-zip files.

See https://github.com/SublimeLinter/SublimeLinter/blob/029ff3dcf441f5606ecb329d66d76ed7df601e96/scripts/repo.json#L14-L14 Over there, I just use "python -m http.server 8000" to host a folder that contains SublimeLinter releases.

That seems unnecessary as the "file:///"-scheme denotes a local file expressed as an url, and frankly, other fetchers would accept them just as-is.  E.g. a simple `urllib.request.urlopen(location)` would work with both http and file urls.

It doesn't work here because of basic tests in `_grab`, namely:

```
        parsed = urlparse(url)
        if not parsed or not parsed.hostname:
            raise DownloaderException('The URL "%s" is malformed' % url)
```

Don't scratch that test but instead check for the "file:///" scheme at the top, and do a straight, dead-simple `file.open` instead.

That's the minimal patch to support the usecase.

### DIFF
--- a/package_control/commands/add_channel_command.py
+++ b/package_control/commands/add_channel_command.py
@@ -30,7 +30,7 @@ class AddChannelCommand(sublime_plugin.ApplicationCommand):
         if isinstance(url, str):
             url = url.strip()
 
-            if re.match(r'https?://', url, re.I) is None:
+            if re.match(r'^(?:file:///|https?://)', url, re.I) is None:
                 output_fn = console_write if unattended else show_error
                 output_fn(
                     '''
--- a/package_control/commands/add_repository_command.py
+++ b/package_control/commands/add_repository_command.py
@@ -30,7 +30,7 @@ class AddRepositoryCommand(sublime_plugin.ApplicationCommand):
         if isinstance(url, str):
             url = url.strip()
 
-            if re.match(r'https?://', url, re.I) is None:
+            if re.match(r'^(?:file:///|https?://)', url, re.I) is None:
                 output_fn = console_write if unattended else show_error
                 output_fn(
                     '''
--- a/package_control/commands/existing_packages_command.py
+++ b/package_control/commands/existing_packages_command.py
@@ -29,7 +29,7 @@ class ExistingPackagesCommand(sublime_plugin.ApplicationCommand):
         default_packages = manager.list_default_packages()
         default_version = 'built-in v' + sublime.version()
 
-        url_pattern = re.compile(r'^https?://')
+        url_pattern = re.compile(r'^(?:file:///|https?://)')
 
         package_list = []
         for package in sorted(self.list_packages(manager), key=lambda s: s.lower()):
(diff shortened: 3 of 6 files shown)


kaste:
Should we test this or do you make releases on a specific schedule?

kaste:
Ah, you made a breaking change here.  You dropped support for absolute paths in "repositories" in the PC settings. I think that change should be additive. It is more consistent now but a breaking change.

deathaxe:
Which bit of lines do you meand?

Absolute local paths are not officially promoted and not supported by _Add Repository_ / _Add Channel Commands_. Supporting them, would also need to add another code path in http_get() for exaclty the same thing. Hence I expect it acceptable for users to adjust possible settings accordingly.

kaste:
> Which bit of lines do you meand?

Just generally, the original patch was +7 additive.  But you could refactor and use it so that we have -61 in the overall diff now.

> Absolute local paths are not officially promoted...

 FichteFoll told me that I can use absolute paths years ago but indeed it is not in the settings comments.

FichteFoll:
Can't you use standard absolute paths in the file schema as well? That's what browsers do anyway. It just looks a bit weird because of the triple forward slashes at the beginning (and I honstly don't know about Windows here).

kaste:
To add support for the file protocol is the meat of this PR.  

In one undocumented part, the PC settings, it was possible to just have standard absolute paths, e.g. `c:\Users\....`.  That is no longer possible.  Users must use the file protocol.

The con is really just that -- as the settings are user facing -- it is easier and feels "normal" to add a standard absolute path, as you can just copy and paste that from everywhere.

deathaxe:
Browsers prepend `file://` when pasting local absolute file paths.

The actual protocol name is `file://`. Third `/` is leading slash indicating absolute local path. Without it, url would denote a relative path, which Package Control expects to be denoted by leading `./`.

Linux: _/www/index.html_ becomes _file:///www.index.html_

Windows: _C:\www\index.html_ is converted to _file:///C:/www/index.html_. The extra `/` being prepended  makes it look like a absolute unix path (to some degree).

PC suppport(ed) adding absolute local paths into "channel" and "repository" settings to support locally hosted channel.json/repository.json by using a separate code path in ChannelProvider and JsonRepositoryProvider.

Any download asset specified via "url" however is downloaded by `http_get()`, which doesn't support local file paths, even if relative urls of releases were resolved to an absolute local file paths.

This made local file hosting impossible.

Of course we could also teach `http_get()` to treat anything not starting with `https://` or `file://` as absolute local file path. But why do we always need hundred ways to do one thing? We'd need to check for https:// vs. file:// vs. local path on all locations changed by this PR.

PR #1697, opened 2025-05-11
Do not cache "file:///" resources

### LINKED ISSUE #1695: Do not cache local repositories or channels
One annoyance with current PC is, that local repositories or channels (but I only ever 
used the former) are still cached.  Basically, making it necessary to reboot
ST whenever you change them manually or programmatically.  (Or you call `clear_cache`
but that doesn't take an argument and thus clears all and everything.)

Of course, the cache can be here for two reasons: (a) save the network call, (b)
save expensive computations after the download.  

If it is only (a) we should just skip the cache for "file://" protocols.

I think the code we need to change is here:

https://github.com/wbond/package_control/blob/20c1cacd13ac8180e5084ca9aad7f0a561c858c2/package_control/package_manager.py#L634-L655

as the caching is not baked (low-level) into the `http_get` function but more high-level when collating the various resources

For the channels, it is in the method just above that.  

If you think that is the correct place to patch, I could give it a try.

kaste:
Fixes #1695

We generally cache to save the network latency, not because our transformations of the data were so expensive.

We expect that "file:///" protocol resources can be read fast enough for our purposes and want to ensure that changes of such local files are available immediately, e.g. without restart Sublime Text or clearing the cache as a whole (`clear_cache`).

Note that we only refuse to read from the cache (`get_cache`) and thus imitate a fresh install.  We still write to the cache in case that is observable by other parts of our source code.

### DIFF
--- a/package_control/package_manager.py
+++ b/package_control/package_manager.py
@@ -493,7 +493,10 @@ class PackageManager:
 
             # Caches various info from channels for performance
             cache_key = channel + '.repositories'
-            channel_repositories = get_cache(cache_key)
+            if channel[:8].lower() == "file:///":
+                channel_repositories = None
+            else:
+                channel_repositories = get_cache(cache_key)
 
             merge_cache_under_settings(self, 'renamed_packages', channel)
             merge_cache_under_settings(self, 'unavailable_packages', channel, list_=True)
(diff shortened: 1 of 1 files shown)


deathaxe:
This changeset just avoids cached values being read. Avoiding values being cached would be preferable.

kaste:
In my last sentence I explain why I did this on the "reading"-side.

> Note that we only refuse to read from the cache (get_cache) and thus imitate a fresh install. We still write to the cache in case that is observable by other parts of our source code.

If you're sure it is safe -- no other code observes our cache files -- we can also do the check on the writing side or both.

kaste:
You squashed this and the commit now has an irritating message:

> Note that we only refuse to read from the cache (`get_cache`) and thus imitate
> a fresh install.  We still write to the cache in case that is observable by
> other parts of our source code.

Maybe, remove that from the commit.  In the future, I'll try to keep the OP of the PR up-to-date.

deathaxe:
Fixed.

PR #1698, opened 2025-05-11
Host main channel.json as GitHub release asset

kaste:
To get hosting "for-free" and to enable ETag and Last-Modified support, cache the official "channel_v3.json" on GitHub.  Ensure to update the release asset only when the contents of the channel actually changed.

It is possible to post-process the "channel.json" in the `generate` phase, i.e. we could drop old ST2 only packages.  But that is not part of this PR.

DISCLAIMER:  This is only useful if our downloaders here actually support ETag/Last-Modified headers.  There is a "caching_downloader.py" but I actually don't know or checked if it is in use here.  

This was less work than one might expect as I just had a the yml action basically at hand with different URL's and names.  So I thought: why not share it?  Maybe it is useful and could solve a tiny problem we have here.  🥂

### DIFF
--- a/.github/workflows/publish_channel.yml
+++ b/.github/workflows/publish_channel.yml
@@ -19,14 +19,14 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
-          python-version: '3.11'
+          python-version: '3.13'
 
       - name: Generate new channel.json
         id: generate
         run: |
-          curl -s -L -o channel.json https://packagecontrol.io/channel_v3.json
+          python scripts/generate_channel.py
           # Verify the file was created
           if [ ! -f "channel.json" ]; then
             echo "Error: channel.json was not generated"
--- a/.github/workflows/publish_channel.yml
+++ b/.github/workflows/publish_channel.yml
@@ -1,0 +1,88 @@
+name: Publish channel.json Asset
+
+on:
+  schedule:
+    - cron: '*/15 * * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+env:
+  RELEASE_TAG: channel-latest
+
+jobs:
+  build-and-publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+
+      - name: Generate new channel.json
+        id: generate
+        run: |
+          curl -s -L -o channel.json https://packagecontrol.io/channel_v3.json
+          # Verify the file was created
+          if [ ! -f "channel.json" ]; then
+            echo "Error: channel.json was not generated"
+            exit 1
+          fi
+
+          # Calculate hash of new file
+          NEW_HASH=$(sha256sum channel.json | cut -d ' ' -f 1)
+          echo "new_hash=$NEW_HASH" >> $GITHUB_OUTPUT
+          echo "New channel.json hash: $NEW_HASH"
+
+      - name: Try to download existing channel.json
+        id: download
+        continue-on-error: true
+        run: |
+          curl -s -L -o existing-channel.json https://github.com/${{ github.repository }}/releases/download/${{ env.RELEASE_TAG }}/channel.json || echo "First run or download failed"
+          if [ -f "existing-channel.json" ]; then
+            EXISTING_HASH=$(sha256sum existing-channel.json | cut -d ' ' -f 1)
+            echo "existing_hash=$EXISTING_HASH" >> $GITHUB_OUTPUT
+            echo "Existing channel.json hash: $EXISTING_HASH"
+          else
+            echo "No existing channel.json found or download failed"
+            echo "existing_hash=none" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Determine if update is needed
+        id: check
+        run: |
+          # If first run or hashes differ, update is needed
+          if [ "${{ steps.download.outputs.existing_hash }}" != "${{ steps.generate.outputs.new_hash }}" ]; then
+            echo "update_needed=true" >> $GITHUB_OUTPUT
+            echo "Update needed: Content has changed or first run"
+          else
+            echo "update_needed=false" >> $GITHUB_OUTPUT
+            echo "No update needed: Content is unchanged"
+          fi
+
+      - name: Update release and upload asset if needed
+        if: steps.check.outputs.update_needed == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Create or update the release
+          gh release view ${{ env.RELEASE_TAG }} || \
+          gh release create ${{ env.RELEASE_TAG }} \
+            --title "Channel Asset" \
+            --notes "Latest channel.json"
+
+          echo "Uploading channel.json file..."
+          gh release upload ${{ env.RELEASE_TAG }} channel.json \
+            --clobber
+
+      - name: Output result
+        run: |
+          if [ "${{ steps.check.outputs.update_needed }}" = "true" ]; then
+            echo "✅ Channel updated with new content"
+          else
+            echo "ℹ️ No changes detected in channel.json, skipped update"
+          fi
--- a/Package Control.sublime-settings
+++ b/Package Control.sublime-settings
@@ -3,11 +3,8 @@
 	// The repositories from these channels are placed in order after the
 	// repositories from the "repositories" setting
 	"channels": [
-		// channel_v4 for python 3.8 compatible libraries to enable plugins
-		// to migrate to python 3.8 until packagecontrol.io supports new scheme.
-		// Note: Must always be located before default channel in the list!
+		// default channel for packages
 		// Repo: https://github.com/packagecontrol/channel
-		"https://packagecontrol.github.io/channel/channel_v4.json",
 		// default channel for packages
 		// Repo: https://github.com/wbond/package_control_channel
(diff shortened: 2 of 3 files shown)


deathaxe:
If we'd like to mirror original channel_v3.json, it would need it would need to be merged into already auto-created channel_v4.json and deployed via existing "GH pages" path.

I am not very sure whether checking file hash of channel_v3.json is enough to check for changes as it is fully re-written after each crawl and contains dictionaries. IIRC packagecontrol still runs on python 3.6 and thus order of dict keys is not ensured to be always the same.

kaste:
> If we'd like to mirror original channel_v3.json, it would need it would need to be merged into already auto-created channel_v4.json and deployed via existing "GH pages" path.

Sorry that I don't understand.  This is meant as a drop-in-replacement for channel3.  Why *need* they to be merged?  They aren't now.  And why deployed on "GH Pages", assets can be attached to releases here as well with no network bandwidth limit basically.  (I actually prefer release assets as user can see them on GitHub which is typical for open source projects.)

Actually, you could just merge with or without the actual settings change to see how the action works, i.e. if it creates a new asset all the time.  If the file gets created from scratch every time, it is also possible that the contents are rather stable and just the Last-Modified date changes all the time.  We can also just try this at home

These are just first tiny steps.  I think we already might win with that no-effort PR here.  We can of course post-process the json or merge two jsons.

PS. Doubt you can crawl 5000 registered packages on Github actions tbh.  Hard to get right with the limited possibilities, no db, no redis, no memcache, no real storage.  You can likely get this for free on Google Cloud Engine however.

kaste:
From my work on the Package x Control add-on I know that mostly everything is developed using git. IIRC only a dozen packages are actually proprietary.  For git based packages, the crawler could still just run `git ls-remote https://github.com/kaste/Package-x-Control.git ...` (whatever the package is) to list all refs or tags.

kaste:
Looks like channel.json isn't stable but bonkers.  Overall, the releases section is ordered randomly.  But also ... packages disappear and reappear kinda randomly.  There seems to be a race between BBCode syntax from https://github.com/Skullmonkey/sublime/blob/master/packages.json#L29 and https://github.com/wbond/package_control_channel/blob/master/repository/b.json#L310 

🤷‍♀️

kaste:
Added very simple script to get the output stable.

deathaxe:
>  This is meant as a drop-in-replacement for channel3.

1. It is not a drop-in replacement, because it requires modification of client settings to update default channel
2. Shipping 2 channels in default settings is already not ideal. Just changing one of them is likely to cause further or avoidable confusion. If everything is deployed via single source (e.g. github) there should (again) only be a single file to download for clients. 
3. GH pages was choosen a) for "nicer" url and b) didn't want to checkin or create releases for each update. Sure, we can update a single existing release's asset each time, but I wasn't aware of that part and GH pages just was the easier and straight forward way back when I set-up this channel.

>  For git based packages, the crawler could still just run git ls-remote https://github.com/kaste/Package-x-Control.git ... (whatever the package is) to list all refs or **tags.**

_That's an idea, I carry around for quite a while, already._

Given that crawling fully relies on `package_control`, which shares functionality with clients (which may not have git installed) this strategy requires to teach package_control to _optionally_ use git backend to get the list of tags and build the internal list(s) of "releases" for each given release branch. 

Package Control currently uses git only to update git cloned packages in a rather naive way. It would require some design changes and infrastructure being added/created.

> There seems to be a race between BBCode syntax

Good finding. This should be resolved. Given that package on package_control_channel is more up-to-date, a way forward was to move external repository's packages and remove it.

>  no db, no redis, no memcache, no real storage. 

Well, that's not entirely true. This channel already stores http cache and most recently crawled data in a cache. If we'd want to build upon a db, I'd built on sqlite moving it to and fetching it from action cache. Without all the stats it would be only few MBs of size. Given those caches are stored for at least a week or longer and crawler running once a day, I think that's doable. With http cache stored locally, amount of counted API calls will quickly shrink to sane amounts as most packages are not changed frequently. The only question is: Do we want to rely on that platform?!

kaste:
> Given that crawling fully relies on package_control, which shares functionality with clients (which may not have git installed) this strategy requires to teach package_control to optionally use git backend to get the list of tags and build the internal list(s) of "releases" for each given release branch.

That is probably what Package x Control is about.  It has these two parts: the dashboard plus a new backend.  The backend relies on git installed and fetches the right tags or branches on what is defined in the registry.   PxC does not install or update packages.  It calls PC for that.  (It would only like to know when theses processes are finished.)

> 

The release assets have download counters for free.  Yeah, artifact and caches, I have here, in this PR, a very simple workflow which is basically testable on the command line.  It doesn't use any other, custom actions, just the gh CLI.  

The platform is probably stable, but if we can't think of a very simple suite of scripts, workflows I don't see an enjoyable dev-cycle.  If it breaks we basically run against a black box, and very typical you just push changes and changes and wait for the runners to fail again.  

It triggers me so I might look at it.

kaste:
In the docs they state 1000 API calls per hour for GitHub actions.  That obviously isn't that much so we need more caching than usual.

kaste:
The script now merges both channels (v3 and the libraries channel v4) into one "channel.json".  

This was a bit counter-intuitive as "usually" to have at least a few resources to download can reduce the download time: usually only a few of them change and needs to be downloaded at a given time.  (So to split the "big" channel in 2 or 4 or per alphabet seemed more intuitive for me.)  However, we don't download in parallel on the client so this is moot and one big download might win.  

Whatever, this merges the channels and publishes them as an asset on GitHub.

kaste:
The script here runs at https://github.com/kaste/pc-e01-thecombine/releases

Interesting that wbond's crawler struggles over the course of 24 hours -- regularly.  Very often SublimeLinter contrib packages disappear, some repositories disappear. And e.g. a package named "npc_sheets" struggles every day.  Ref: https://github.com/kaste/pc-e01-thecombine/actions/runs/15306380956/job/43060079641 (removed) 
 https://github.com/kaste/pc-e01-thecombine/actions/runs/15306860765/job/43061650887 (added again). Or: https://github.com/kaste/pc-e01-thecombine/actions/runs/15303234565/job/43049136470

Interestingly, always the same packages.  Roughly at the same time.

A new crawler runs at https://github.com/kaste/pc-e02-thecrawl/releases

deathaxe:
> That would be very transparent, but wbond is gone. 

That's my point. All repos under wbond/ are basically locked, just with push access for small number of users and so is this one here. So whether to use package_control or package_control_channel for deployment doesn't make a difference, besides putting it here mixing things up!

The only truely open location would be SublimeText or packagecontrol organization, which are in community hands.

TBH, once packagecontrol/channel would be able to live on its own, I'd even be open to make it the primary source of channel data and let current ecosystem die.

> However, I don't get the pro of the stable, or keep the current v4 URL intact argument.

There were enough reports and support efforts to get settings fixed for those who customized them and thus missed changed defaults, when we introduced v4 channel for libraries.

There's a way around doing so again, and this is to take.

As channel data are no packages, but just a regularly updated index, I don't get the "it must be a github release asset" argument.


> Ideally, wbond would transfer the channel repo to a trusted org.

Sure.

If we wouldn't need to maintain packagecontrol.io we could just fork them into github/packagecontrol organization.

> Or he's just a psycho with a control issue

I guess this is just the way to say he doesn't care about the ecosystem and its future anymore and has absolutely no interest in putting any but least effort into. It just says: make something new or live with what you have.

kaste:
> There were enough reports and support efforts to get settings fixed for those who customized them and thus missed changed defaults, when we introduced v4 channel for libraries.
>
> There's a way around doing so again, and this is to take.
>
> As channel data are no packages, but just a regularly updated index, I don't get the "it must be a github release asset" argument.

I think, it is not the URL but the website and tooling around it that is compelling.  The tooling: I can just use the gh CLI on the action runner and from home to download and view assets.  And the website of course: there is a website, the releases page with its information and links to the runner log, the sha, the date.  I think it is more transparent and trustworthy if I don't have just one link/URL to download the channel from but a page where this link is embedded in, and which I can lookup.  The wbond channel also doesn't do this, and I missed it then as well.  But that is really a matter of taste: IMO for a small open source community it should be KISS and use the infrastructure others built for us.  Recreating a website around a asset URL is more work.

The migration pain path is for sure an argument, but a) we would ship a message.txt on the update, *and* b) when we read the channel we tell (or even fix) the wrong URL in their settings.  (We "validate" it or actively migrate the settings.)   That is less work here in code, than for *all* the users to ask for it.   

We probably need a group on discord to discuss the path forward.  At least braver is required.  I think Terminal also had the email to contact wbond, but I don't remember.  I remember wbond saying that he wants to be emailed(?) again if the (old) crawler stops again.   (Very vague.)

deathaxe:
Github deploys all release assets with content type `application/octet-stream`. As Package Control excludes this type of content from HTTP cache as this is used for normal package ZIP files, using release assets to deploy channel is ruled out.

kaste:
(the download_manager is really a bastard of old, legacy code. 😁)

deathaxe:
Well, that's actually also how modern httpx based caching modules work, except those not supporting type based exclusions, currently.

deathaxe:
Prefer_cached is unrelated here. It just forces to use an already cached value without reaching out to server to verify it being up-to-date.

kaste:
Yes, I know.  I don't think prefer_cache is unrelated.  http_get is the public function we should use,  and if I set it to True here it might get ignored downtown silently.  The reader and user of http_get can't know that.  That's what I mean with "factoring".  

Anyhow, that's just not where we are right now.  You can of course make a nice website.  But we don't have any right now.  The first commit is not the final product.  And the first commit follows the KISS principal.  (In the long run we need to make a website from the channel, as replacement for packecontrol.io, anyway and anyhow.)

kaste:
I'll close this here as it has obviously been superseded although interesting bits of information are still here in the thread.

PR #1701, opened 2025-05-21
Safely convert file uris to local paths

kaste:
The previous `url[8:]` is wrong as it strips the leading "/" on *nix filesystems.

On modern python, we would use `Path.from_uri(url)` here, so let's backport that.

We then see that we also forgot the important `is_absolute` check.

### DIFF
--- a/package_control/download_manager.py
+++ b/package_control/download_manager.py
@@ -1,5 +1,4 @@
 import os
-from pathlib import Path
 import re
 import socket
 import sys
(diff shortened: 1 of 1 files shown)


package_control/download_manager.py
```diff
@@ -1,8 +1,10 @@
+import os
+from pathlib import Path
```
deathaxe, line 2: pathlib is not a possible dependency to rely on in Package Control 4.x as it is not available on python 3.3. Vendoring for just this single task is not desired.

PR #1703, opened 2025-05-30
Show already installed package in the "Install Package" panel

### LINKED ISSUE #1569: Include installed packages in the "Install Package" quick panel
There have been numerous reports of people trying to install a package and not being able to find it from within PC inside ST because they already had it installed but ignored. 

It would be useful if PC listed these package in the "Install Packages" quick panel regardless but annotate them with a "[installed]" tag or similar to signal that this package will not be installed again.

Additionally, PC could add a tag for "[disabled]" packages for even more clarity.

kaste:
Fixes #1569

Users tend to forget what they installed or installed and then disabled.

Include in the "Install Package" panel all these packages as well. For the disabled packages, rewrite the action and just enable the package.


kaste:
"Bracket Color Scheme" is not disabled in your case, or is it?  I think the code is straight forward and correct in what it considers to be disabled.

```
disabled = installed & load_list_setting(settings, 'ignored_packages')
```

kaste:
Yes, better don't hit `enter`.  😁   You can do a noop of course.  The UI demands for a verb in that position.  

But then I thought: okay people are maybe searching for packages that they don't think they have already installed because they actually *don't* work.  I get regular user reports that users just re-install packages if they don't work.  That's where this decision comes from.  It is still an action panel.

I do like davids mockup he showed on discord.  He also provided a "reinstall package" for installed packages but then some more.  IIRC for installed packages it open a next panel with reinstall, uninstall, show homepage, show readme or so.  But that's a whole other level.

kaste:
This is just a baby-step again and basically in style of what we already have.  

To get it "right" in one step is basically too much to ask if everything at every edge is on fire.  Remembers me when I took GitSavvy (and you made a simple PR how to optimize the find_git_root thing -- took me a year to actually implement this), everything was just loosely working, and every commit had a "but..." somewhere.

I didn't even know that we still ship new versions to ST3 because there is ".python-version" at the root that misguided me.  (I thought parts of the software were copied to the py33 host on the fly.)   Why is that?  Sure ST3 people get any updates from the channel.json, but they can't expect new versions of PC! (?)  (The plan for -- or: the ability of: -- the "new crawler" was to write st3 and st4 channels separately, we already save 1MB for the st4 channel, and the st3 channel might save too or naturally dies and turns into a static asset over time.

Me:
> I do like davids mockup he showed on discord.
You:
> Someone recently even made a suggestion (on Discord?)

That should be the same thing.  We should just ask him for the code.  Or maybe that is an aspect you really like to program in Sublime Text  -- (I don't - I esp. hate the design aspect of it.)

That being said switching off py33 is not a major step.  Nothing changes for the ST3 users and nothing changes for the ST4 users.  (I never do a major when switching off an old python version.)  We can just make it, freeze a st3-4.0.9 and then do the other baby-steps.  Obviously, wbond is not here so we do need an override of Package Control in the main registry (p_c_channel) because that is registered in wbond private repo.  🙄

PR #1704, opened 2025-05-31
Fix failed actions to get reported as "successfully"

kaste:
The conditional `num_packages == 1` clearly reports a success even if the action failed as it does not consider failures.

Fix that and add a clear call to action if any failures occurred. Typically, restarting Sublime Text heals every wound.

### DIFF
--- a/package_control/package_tasks.py
+++ b/package_control/package_tasks.py
@@ -395,14 +395,18 @@ class PackageTaskRunner(PackageDisabler):
                 elif result is None:
                     package_names.remove(task.package_name)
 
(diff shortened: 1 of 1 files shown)


kaste:
From an UI perspective, the ActivityIndicator should run for another 2 ror 3 seconds after `__exit__` being called.  (Basically the timing of a window.status_message.)  It looks like the final messages disappear too soon.  (Lucky us, we have the wait time of 0.7secs here.)

deathaxe (approved):

deathaxe:
Nice finding.

PR #1705, opened 2025-06-11
Fetch unavailable libraries directly from pypi

kaste:
If a requested library is not available in one of the repositories,
usually the channels "libraries_cache", fetch the information directly
from PyPi.  

The core-feature is to automatically discover suitable versions of a 
library, even when the Python host changes or gets compiled differently. 

Vendors [packaging](https://github.com/pypa/packaging).  Requires the 3.8 host.

This should e.g. fetch 313 automatically if a new host arrives without changing the libraries cache.


deathaxe:
The info is certainly of value and is very likely to be used as installing libraries directly from pypi without maintaining a dedicated index is certainly a primary goal for the next major release of PC, but combined with proper recursive dependency resolving. 

I am just not yet at the point to say PC5.x will support all ST4 builds including those with py38 or whether we'll hit a point of requiring it py3.13+, leaving all users with current ST4 releases behind with PC4.x

Thus any upstream ecosystem may need to be prepared to handle that.

kaste:
Yes, totally correct.  This PR did look for the libraries first, and then had the fallback to fetch directly.  I hope I did that.  That kind of hybrid approach is a must.  For performance as well.  Gatekeeping is also important, there are actually so mixed emotions for this one.

a) I just wanted to ensure or saw a way how 3.13 dependencies could just work without the need for stressful actions on our (yours) side.  
b) I saw finally a way to install a lib *easily* (it was possible before of course) on my own PC; it is my computer, power to the people.  Gate-keeping would be then up to package reviews if a user then would like to have that on the public tracker.  But my personal stuff, I don't want that on the public tracker.  

(b) is basically a user story.  It is my computer, can PC help with installing libraries on it  

Maybe maintaining allow- and deny-lists would be an option.

PR #1706, opened 2025-07-01
New stats for https://packages.sublimetext.io/

deathaxe:


package_control/package_manager.py
```diff
@@ -2326,22 +2327,44 @@ def record_usage(self, params):
...
+                    '''
+                    Error submitting usage information for %s
+                    ''',
+                    params['package']
+                )
+
+        # new stats
+        url = self.settings.get('submit_usage_url', '')
+        if url:
+            url += '?' + urlencode({"pkg": params["package"], "type": params["operation"]})
```
FichteFoll, line 2356: Afaict, the new stat reporting includes strictly less information than the old one (only two parameters with a different name). Was there any particular reason why the old interface was not adopted, @kaste?  The only one I can think of would be reducing the amount of information that is transmitted to the absolute minimum for data protection, but especially things like the ST and PC version seem quite useful for determining the target audience. PC has historically been our only place to collect and provide this kind of information.

If the new endpoint used the same interface, we could simply make the `submit_url` an array and have it report to the two (or any number of) URLs with all the available data and then just ignore at the server side what we aren't interested in (yet).

FichteFoll:
If we're willing to change the interface further, We could also look into batching stat reports when multiple package operations are performed at the same time to save a few requests and bandwidth, since I imagine most updates to be performed on ST start.

deathaxe:
Batching may certainly make sense as long as requests are sync. Not sure how important it will be with httpx sending requests via HTTP2 though. If those are just fired they should go through one connection one after another anyway without noticible delays. We don't need to wait for them to return tu continue installing next packages.

package_control/package_manager.py
kaste: I think wbond overshared data.  We only wanted to collect data that we also "give back" to the user -- in the UI of the website.  That makes it transparent.  I think there must be a plan how we present the data.  And if that justifies the collection.  Data is always interesting.  We did not even use the update/removed event type as of yet. wbond just showed the ever increasing "install" count.

kaste:
We don't have a bandwith/requests issue here at all.  That would have been the case if we would store each event as a row/entry, and then do views/aggregates on the SQL and so on.  With the current design of the db, that cannot be the case.

deathaxe:
This request probably targets client side delays caused by sending pings synchronously. It may cause some delays when batch install/upgrade many packages.

package_control/package_manager.py
kaste: As I did this, OS specific stats were dropped as "not-planned" UI wise.  Maybe the PC release cycle should allow some spontaneity.

TerminalFi: @FichteFoll send the data, we just drop unused data for now. But if you send it, we can still use it later as desired

deathaxe:
Array doesn't work as both end points use differend parameter names as well as response bodies. However I actually don't care about it being an array or not as the old one will be dropped as soon as packages.sublimetext.io matured enough.

PR #1714, opened 2025-10-29
Prepare for python 3.14

deathaxe:
Just in case ...

### DIFF
--- a/package_control/distinfo.py
+++ b/package_control/distinfo.py
@@ -157,7 +157,7 @@ class DistInfoDir:
             specific to a platform and optionally architecture
         """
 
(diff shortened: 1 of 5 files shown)


michaelblyons:
Did 42xx get pushed from 3.13 to 3.14 for people actively using it, or is this still from comments that the next anticipated Dev channel release will be 3.14?

deathaxe:
This is just fixing existing path name issues and preparing for a smooth transition to 3.14 for those using private beta builds. No it is not available at the time of writing this.

Note: 7e3f00bddbf5372e1571e8e57a2241bc0f6cef74 and 2c062e797e9854916a00e3791acc5a5c659a3cc1 also belong to those considerations. Just forgot to push those commits to this PR before merging.

